PR #25, opened 2017-10-10
🐛 Fixed getState priority

omenocal:
When using this method:
```
app.toStateIntent('anotherState', 'anotherIntent');
```
The new state is saved in the `response (AlexaResponse)` object, however, when calling the `app.getState()` method, it first looks for the old state in the request when I already changed it in the response to the new one.


jankoenig:
Thank you! 🙌

PR #42, opened 2017-11-09
Fixed table markup

botmaker-florian:
Hi Alex and Jan,
this is just the smallest contribution ever, but upon reading the documentation I saw this list that wanted to be a table, and quickly fixed it.
Kind regards
Florian


jankoenig:
Thank you!! 🎉

PR #53, opened 2017-12-21
Fix examples link in examples readme

frozzare:
`./examples/indexRenderTemplate.js` will add `examples` twice so removing `examples` will fix the links :)


jankoenig:
Thanks a lot, @frozzare 👍

PR #151, opened 2018-05-16
Add <phoneme> support to SpeechBuilder

fgnass:
This PR adds support for `<phoneme>` tags to the SpeechBuilder class.
Since Google does not support phonemes the GoogleActionSpeechBuilder adds the text _as-is_.


aswetlow:
Thanks!

PR #169, opened 2018-06-08
Prevent (and gracefully handle) DynamoDB errors

fgnass:
## Proposed changes

DynamoDB does not support empty Strings. If the session contains empty (or `null`) values one gets the following error upon save:

> ValidationException: One or more parameter values were invalid: An AttributeValue may not contain an empty string

The error is not properly handled and yields an unhandled promise rejection error.

This PR addresses both issues and emits a new app-event called `responseError` to allow plugins to report/log the error. 

## Types of changes

## Checklist


aswetlow:
Thanks Felix!

PR #205, opened 2018-07-25
fix body template 6

vitaliidanchul:
for body template 6 we need to setBackgroundImage instead setImage.
Testing this with echo Spot, and echo Show - it works!


aswetlow:
Thanks @vitaliidanchul !

PR #216, opened 2018-08-03
Update Skill Events docs including new Staging section

rmtuckerphx:
Added to existing docs to show for each event what data (userId, permissions) is available and how to access it. Showed how to delete a user when a skill is disabled. Added section on Staging to show how different stages can set the endpoint to handle skill events.


jankoenig:
Very cool, thank you!

PR #227, opened 2018-08-14
Alexa - AudioPlayer - add PlaybackController events

kouz75:
## Proposed changes
Add compatibility to GUI device (Echo show/Spot)
If user press previous/next/resume on Echo Spot device, framework will crash :
TypeError: Cannot read property 'type' of undefined
    at Promise (node_modules/jovo-framework/lib/handler.js:33:23)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [x] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #229, opened 2018-08-14
Feature/playback controller

vitaliidanchul:
Implementing playback controller


jankoenig:
Wow, this has never happened before: Basically the same pull request by two different people on the same day! (see #227).

Thanks a lot for this. We will take a closer look and see how we can merge it with the other contribution. 👍 Cool stuff

vitaliidanchul:
Oh, didn't see that) #227 will not send response, if we use `this.alexaSkill().audioPlayer().setOffsetInMilliseconds(this.user().data.offset).play(song);`

aswetlow:
Thanks @vitaliidanchul 
Yeah, you are right. This part was missing: ` || this.jovo.isPlaybackControllerRequest()`

`PlaybackController.PauseCommandIssued` never gets fired (tested on my Android phone). Does that work for you?

vitaliidanchul:
@aswetlow , no it doesn't work for me, too. some devices handle pause locally:
image

PR #263, opened 2018-10-10
Fix exception when an error message is returned from Alexa

theBenForce:
## Proposed changes
When the Alexa service returns an error message, Jovo throws an exception because the error request object doesn't have the getSessionAttributes method. This change adds a check for that method before calling it.

## Types of changes

## Checklist


theBenForce:
I'll clean this up tonight and move the second commit to its own pull request.

For some background, these are both issues that show up when trying to send different animations to different buttons.

PR #314, opened 2018-12-31
Fix for intent exception causing unhandled rejection

IGx89:
## Proposed changes
This fixes an issue I encountered where the app would hang (and log an unhandled exception) if an intent immediately threw an exception. The root cause was using an async function as a Promise executor (more details here: https://eslint.org/docs/rules/no-async-promise-executor). There's no tslint rule yet for it unfortunately.

## Types of changes

## Checklist


aswetlow:
Thank you @IGx89

PR #366, opened 2019-01-25
Update response.md

VoiceFirstTech:
## Proposed changes

the docs state .getSpeech() does not return <speak> tags 
However, .getSpeech() does return <speak> tags 

image

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x ] My code follows the code style of this project
- [ x] My change requires a change to the documentation
- [x ] I have updated the documentation accordingly

### DIFF
--- a/docs/basic-concepts/requests-responses/response.md
+++ b/docs/basic-concepts/requests-responses/response.md
@@ -38,8 +38,8 @@ You can use these cross-platform helper methods:
 
 | Method        | Description        | 
 | ------------- |-------------| 
-| `getSpeech(): string` | Returns speech text without "speak" and tags. | 
-| `getReprompt(): string` | Returns reprompt text without "speak" tags. | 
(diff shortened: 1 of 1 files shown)


jankoenig:
Hi there, thank you! We think this is rather a bug in the code than a docs issue. We're working on updating the code so that it returns the speech withouth the `speak` tags. What do you think?

VoiceFirstTech:
Perfect! 
 <speak> Let's get rid of those tags </speak>

we can close the issue if the change is in the backlog :+1:

jankoenig:
Let's keep this open so we stay motivated to fix this fast

jankoenig:
Fixed with the latest release! https://github.com/jovotech/jovo-framework/blob/master/CHANGELOG.md#2013-2019-01-29

Issue #431, opened 2019-03-22
Expanded support for SSML


jnthnwn:
## I'm submitting a...

- [ ] Bug report
- [x] Feature request
- [ ] Documentation issue or request
- [ ] Other... Please describe:

## Expected Behavior

There should be support for additional elements of [SSML](https://developers.google.com/actions/reference/ssml), as far as it's supported by either platform.

Certain sentences, words, or possibly sessions should have a programmatic way to add elements such as `prosody` or `sub` to better describe how to speak.

For example:
```
// for an entire sentence
let prosody = { rate: 'slow', pitch: '-2st' };
this.$speech.addText('hello world');
this.$speech.prosody(prosody);
this.tell(this.$speech);

// or for part of a sentence:
this.$speech.addText('hello', { prosody });
this.$speech.addText('world');
this.tell(this.$speech);
```

jankoenig:
Thank you. Yes, this is a great suggestion. I think something like this would also be a great opportunity for a community contribution/pull request if anyone wants to help 💪

aswetlow:
I close this issue due to inactivity.

PR #442, opened 2019-03-28
Update addText to accept arbitrary SSML wrapping elements

jnthnwn:
## Proposed changes
Changes to SpeechBuilder's addText method to support arbitrary SSML elements. #431 

These changes simplify adding other helper methods for SSML syntax and allow users to add SSML supported by a specific platform even if Jovo hasn't been updated its API with helper methods.

I'd like feedback on if it would make sense for another argument to added to the other public methods as well, so a user could addSentence with some surrounding SSML tags.

## Types of changes

## Checklist

- [X] I have added tests to cover my changes

### DIFF
--- a/jovo-core/src/SpeechBuilder.ts
+++ b/jovo-core/src/SpeechBuilder.ts
@@ -1,9 +1,19 @@
 'use strict';
 import {Jovo} from "./Jovo";
 const _sample = require('lodash.sample');
+const _merge = require('lodash.merge');
+
+export interface SsmlElements {
+    [tag: string]: SsmlElementAttributes;
+}
+
+export interface SsmlElementAttributes {
+    [attribute: string] : string | number | boolean;
+}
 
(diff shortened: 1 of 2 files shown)


jankoenig:
Thank you!
We will take a deeper look early next week and provide feedback :)

Have a great weekend,
Jan

aswetlow:
Great PR! Thanks
Any idea of how to solve "self-closing" tags like `break` or `audio`?

I also like your idea about the `addSentence` method. Feel free to add more functions :)

jnthnwn:
added [this line](https://github.com/jovotech/jovo-framework/pull/442/files#diff-fdae62756d94c2f63bac9578691b8e45R205) to address self-closing tags.

i have added the surroundSsml argument to some of the convenience methods, but i think that adding more convenience methods might be a hassle to maintain later on. i think the changes to addText should be sufficient syntax for users to modify the speech.

PR #477, opened 2019-04-18
Fix dialogflow test suite

kozikkam:
## Proposed changes
This PR makes dialogflow test suite work. Previously, DialogflowResponseBuilder had two class attributes: platform and platformResponseClazz. Both were undefined. Additionally, DialogflowResponse has optional attribute `outputContexts?: Context[]` which then is used like so:
```
const sessionContext =_get(this, 'outputContexts').find((context: Context) => {
```
causing an exception to be thrown, if it's undefined.

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/jovo-integrations/jovo-platform-dialogflow/src/core/DialogflowResponse.ts
+++ b/jovo-integrations/jovo-platform-dialogflow/src/core/DialogflowResponse.ts
@@ -52,6 +52,9 @@ export class DialogflowResponse implements JovoResponse {
 
 
     getSessionAttributes() {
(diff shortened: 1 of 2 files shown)


jankoenig:
Hi @kozikkam, thank you for this pull request. Just noticed you closed it. Any reason for that?

Happy to help,
Jan

kozikkam:
@jankoenig The issue is deeper than I thought, I will open an issue.

PR #479, opened 2019-04-23
Add google res getters

natrixx:
## Proposed changes
add the following initial google specific response getters:

hasSuggestionChips()
getSuggestionChips()
hasDisplayText()
getDisplayText()

I plan to add more, but want to make sure I am on the right track, and keep the PR's manageable :) 

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [x ] My change requires a change to the documentation
- [ x] I have updated the documentation accordingly
- [x ] I have added tests to cover my changes
- [x ] All new and existing tests passed


aswetlow:
Thank you, Nate!

There's one thing I would edit. We can use up to 8 suggestion chips. We need to make the `hasSuggestionChips(chipOne?: string, chipTwo?: string, chipThree?: string):` more flexible.

Here's a code sample what I have in mind. (barely tested)
```typescript
hasSuggestionChips(...chips: string[]){
    const suggestionChipArray = this.getSuggestionChips();

    if (!suggestionChipArray) {
        return false;
    }
    for (let i = 0; i < chips.length; i++) {
        if (!suggestionChipArray[i] || chips[i] !== suggestionChipArray[i].title) {
            return false;
        }
    }

    return true;
}

```

natrixx:
ooh good catch, I remember seeing that long ago and far way, but had forgotten. 

I will work on this and see how it goes, thanks!

natrixx:
Ok @aswetlow - that suggestion appears work great! I added a couple more tests to include a full array of options. I think that along with the previous tests with combos of up to 3 items in the array should be enough coverage. Thoughts?

aswetlow:
Perfect! Thank you

PR #525, opened 2019-06-13
Fixing a bug in server.js that breaks the express server

daveradical:
I upgraded to the latest version of jovo and suddenly nothing worked so I did some investigating and realised there was a mistake in the server.js. It was using the non-verified server in the verified server listen function so nothing with express worked.

## Types of changes

## Checklist


aswetlow:
Thank you! I overlooked that.

PR #545, opened 2019-07-19
Fix: Prevent validators to fail after redirect

gpalozzi:
## Proposed changes
Pass the input name being validated by a validator to the `parseForValidator` function.
Before this change, the function was raising an error due to the lack of `$input` parameters which cannot be passed via an intent redirect.

https://github.com/jovotech/jovo-framework/blob/0558d48d33779f3f560175e7ae20501b0f15540e/jovo-core/src/Jovo.ts#L653

Even if an intent is defined to receive an input parameter in the model definition,
it might be called via a redirect (e.g. `toIntent()`) which won't populate the `$input` object.

Therefore, it won't be possible to retrieve the input name from the `$input` object,
but it is available within the validation schema definition.

## Types of changes

## Checklist

### DIFF
--- a/jovo-core/src/Jovo.ts
+++ b/jovo-core/src/Jovo.ts
@@ -552,6 +552,7 @@ export abstract class Jovo extends EventEmitter {
                 for (const v of validator) {
                     await this.parseForValidatorAsync(
                         v,
+                        input,
                         this.$inputs[ input ],
                         failedValidators,
                     );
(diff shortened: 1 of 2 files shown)


aswetlow:
Hey @gpalozzi 

Thank you for the fix! Let me add @rubenaeg to this thread who implemented this feature.

PR #549, opened 2019-07-25
:sparkles: Added support for random hints

frivas:
With this change it is possible to use an array of hints that will be shown using the `showHint` function like it is done with `addText`. Now this function accepts a string or an array of strings, this way extra dynamism is added to the Skill.

## Proposed changes

The declaration and definition of the function `showHint` had been changed to support either a simple string or an array of strings. The implementation had been taken from the `addText` one.

## Types of changes

## Checklist

*Notes: Seems there are no previous tests for the current `showHint` implementation. Also seems that I can't update the documentation myself. Please, advice and thank you very much in advanced.*


aswetlow:
Thank you!

PR #575, opened 2019-09-13
🐛 Fix isNewSession for Dialogflow platform

### LINKED ISSUE #536: Dialogflow Agent isNewSession Not Implemented
## I'm submitting a...

- [ ] Bug report
- [X] Feature request

## Expected Behavior
Calling isNewSession on the DialogflowAgent should return True if a new session, or False otherwise. There is nothing in the dialogflow request that explicitly states its a new session, however it could probably be implicitly determined when there are no output contexts, or something else? Something more complicated than just reading a new session field from the request needs to be implemented.

## Current Behavior
The isNewSession implementation in DialogflowAgent.ts line 52, currently is hard coded to return True every time.

gpalozzi:
## Proposed changes
Currently the `isNewSession` method of a Dialogflow Agent is returning `true` for every request.

This fix will actually inspect the `outputContext` of the request to check whether the `_jovo_session_` context is present.
This will happen only when the `payload` doesn't contain an `isNewSession` method itself.

The fix should resolve #536 

## Types of changes

## Checklist

### DIFF
--- a/jovo-platforms/jovo-platform-dialogflow/src/DialogflowAgent.ts
+++ b/jovo-platforms/jovo-platform-dialogflow/src/DialogflowAgent.ts
@@ -50,7 +50,7 @@ export class DialogflowAgent extends Jovo {
      * @return {boolean}
      */
     isNewSession(): boolean {
-        return true;
+        return this.$request!.isNewSession();
     }
 
 
--- a/jovo-platforms/jovo-platform-dialogflow/src/DialogflowCore.ts
+++ b/jovo-platforms/jovo-platform-dialogflow/src/DialogflowCore.ts
@@ -124,6 +124,15 @@ export class DialogflowCore implements Plugin {
             return !context.name.startsWith(sessionContextPrefix) && context.lifespanCount && context.lifespanCount > 0;
         });
 
+        // add ask context
+        if (output.ask) {
+            newOutputContexts.push({
+                name: `${request.session}/contexts/_jovo_ask_${Util.randomStr(5)}`,
+                lifespanCount: 1,
+                parameters: { ask: true }
+            });
+        }
+
         // add jovo session context
         if (Object.keys(dialogflowAgent.$session.$data).length > 0) {
             newOutputContexts.push({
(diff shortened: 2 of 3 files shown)


aswetlow:
Hey @gpalozzi 
Thank you for the PR.

That will solve the issue for now. But we probably need to set the _jovo_session_ in the Dialogflow implementation implicitly in `ask` responses, which don't set any other session attributes. Otherwise `this.getSessionContext()` will always be undefined.

Here's an example:

```

app.setHandler({
    LAUNCH() {
 
        console.log('Initial start: ' + this.isNewSession());
        this.ask('Hello World! What\'s your name?', 'Please tell me your name.');

    },

    MyNameIsIntent() {
        console.log('Second request: ' + this.isNewSession());
        this.tell('Hey ' + this.$inputs.name.value + ', nice to meet you!');
    },
});
```
Conversation
```
Talk to test app
- What's your name?
My Name is Jeff
- Hey Jeff, nice to meet you!
```

Both `this.isNewSession()` are true in this case because there is no context in both requests.
Something like `this.$session.$data.ask = true` will add the context object...

What do you think?

gpalozzi:
I think you are definitely right @aswetlow. We should add at least one element to the session in order to create the context object. I'll have a look ;)

aswetlow:
I played around. What about a second independent context for ask responses that aren't affected by the other `this.$session.data` objects. 

Here's the code:


Add a _jovo_ask context in `output()` (DialogflowCore.ts)
```
 if(output.ask) {
            newOutputContexts.push({
                name: `${request.session}/contexts/_jovo_ask_${Util.randomStr(5)}`,
                lifespanCount: 1,
                parameters: {ask: true}
            });
 }
```

Add `getAskContext()` helper method (like `getSessionContext()`)  (DialogflowRequest.ts)
```
getAskContext(): Context | undefined {
        const sessionId = this.session;

        if (this.queryResult && this.queryResult.outputContexts) {
            return this.queryResult.outputContexts.find((context: Context) => {
                return context.name.startsWith(`${sessionId}/contexts/_jovo_ask_`);
            });
        }
    }
```

Look for the context in `isNewSession()` (DialogflowRequest.ts)
```
isNewSession() {
        if (typeof _get(this.originalDetectIntentRequest, 'payload.isNewSession') === 'function') {
            return this.originalDetectIntentRequest.payload.isNewSession();
        }
        const askContext: Context | undefined = this.getAskContext();
        return typeof askContext === 'undefined';
    }
```

gpalozzi:
👍 I'll give it a try!

aswetlow:
Perfect. Thank you!

Issue #583, opened 2019-09-22
Jovo Debugger frontend can only parse empty self-closing audio SSML tag


tuffnatty:
## I'm submitting a...

- [x] Bug report
- [ ] Feature request
- [ ] Documentation issue or request
- [ ] Other... Please describe:

## Expected Behavior
Jovo Debugger shows a play button when it receives an `<audio>` SSML tag, no matter whether the tag is empty and self-closed (`<audio src="..."/>`) or not (`<audio src="..."></audio>`).

## Current Behavior
Jovo Debugger shows a play button only for empty self-closed `<audio>` SSML tags.

* Jovo Framework version used: 2.x.x

PR #584, opened 2019-09-22
Fix GoogleActionSpeechBuilder.addAudio() to match docs+implementation+common sense

tuffnatty:
## Proposed changes
This changeset adds tests for GoogleActionSpeechBuilder and fixes the following issues:
* wrong type declaration on url parameter
* when text is not given, assume empty text, don't substitute it by url
* crash when `url` is an array but `text` is not
* corrupt output when `url` is not an array but `text` is
* when text is empty, output XML-compliant self-closing tag (but this will mask away bug #583) as AlexaSpeechBuilder already does.

## Types of changes

## Checklist


aswetlow:
Thank you!

PR #607, opened 2019-10-28
fixed removing shouldEndSession in showVideo call

BronxBombers:
## Proposed changes
Update showVideo method to properly remove shouldEndSession property.
Since shouldEndSession is a boolean property, when the value was False it was not being removed from the output. Alexa does not accept shouldEndSession appearing at all on video responses.

## Types of changes

## Checklist

- [x ] My code follows the code style of this project
- [x ] All new and existing tests passed


aswetlow:
Thank you!

PR #609, opened 2019-10-30
💥✨Adjusted State Handling to prevent leaving active Conversational Components

stammbach:
## Proposed changes
Added "active component" checks to `toStateIntent`, `toStatelessIntent` & `followUpState` to prevent leaving the conversational component handler when using those "state" related methods.

## Types of changes

_Not sure if it can really be called a breaking change because component are still in development_

## Checklist

I don't think so, because from a user standpoint I assumed to stay inside my component when using states. 
All existing test are running fine for me

### DIFF
--- a/jovo-framework/src/middleware/Handler.ts
+++ b/jovo-framework/src/middleware/Handler.ts
@@ -369,6 +369,8 @@ export class Handler implements Plugin {
                 : this.getRoute().path;
             this.$plugins.Router.route = route;
             Log.verbose(` toStateIntent: ${state}.${intent}`);
+
+            this.triggerStateValidation = true;
             return Handler.applyHandle(this, route, true);
         };
 
--- a/jovo-framework/src/middleware/Handler.ts
+++ b/jovo-framework/src/middleware/Handler.ts
@@ -14,7 +14,7 @@ import _get = require('lodash.get');
 
 
 import { Config as AppConfig } from './../App';
-import { ComponentDelegationOptions, ComponentResponse, ComponentSessionData } from './Component';
+import { Component, ComponentDelegationOptions, ComponentResponse, ComponentSessionData } from './Component';
 import { ComponentPlugin } from './ComponentPlugin';
 import { Route, Router } from './Router';
 
@@ -335,23 +335,52 @@ export class Handler implements Plugin {
             state: string | undefined,
             intent: string,
         ): Promise<any> {
+            // Check for Component Names inside State
+            if (typeof state !== 'undefined' && this.$components) {
+                const components: string[] = Object.keys(this.$components);
+                const states: string[] = state.split('.');
+                const matched: string[] = states.filter(s => components.includes(s));
+
+                if (states.length > 1 && matched.length > 0) {
+                    throw new JovoError(
+                        `Can not use component names as states. Please rename the following states: ${matched.join(', ')}`,
+                        ErrorCode.ERR,
+                        'jovo-framework',
+                        'The state you want to use contains at least one component name.',
+                        'Rename the listed states to not interfere with them.'
+                    );
+                }
+            }
+
+            const currentState = this.getRootState();
+            if (typeof currentState !== 'undefined' && typeof state !== 'undefined') {
+                let currentStates: string[] = currentState.split('.');
+
+                if (typeof 'state' !== undefined && currentStates[currentStates.length - 1] !== state) {
+                    currentStates = [ ...currentStates, state];
+                }
+
+                state = currentStates.join('.');
+            }
+
+            return this.internalToStateIntent(state, intent);
+        };
+
+        /**
+         * Jumps to state intent without any validation
+         * @private
+         * @param {string} state name of state
+         * @param {string} intent name of intent
+         */
+        Jovo.prototype.internalToStateIntent = async function (
+            state: string | undefined,
+            intent: string,
+        ): Promise<any> {
             // tslint:disable-line
             this.triggeredToIntent = true;
 
-            const componentSessionStack: Array<[string, ComponentSessionData]> = this.$session.$data[SessionConstants.COMPONENT];
-            if(componentSessionStack != null && componentSessionStack.length > 0) {
-                const activeComponentSessionData: [string, ComponentSessionData] = componentSessionStack[componentSessionStack.length - 1];
-                const activeComponent = this.$components[activeComponentSessionData[0]];
-
-                if(state !== activeComponent.name) {
-                    Log.verbose(` Changing component state to: ${state}`);
-                    state =  `${activeComponent.name}.${state}`;
-                }
-            } else {
-                Log.verbose(` Changing state to: ${state}`);
-            }
-
             this.setState(state);
+            Log.verbose(` Changing state to: ${state}`);
 
             const route = Router.intentRoute(
                 this.$handlers,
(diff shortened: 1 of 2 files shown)


stammbach:
Hey @KaanKC,
haven't thought about the nested components, sorry! But then there is another problem, because I basically copied the logic from the `sendComponentResponse` method which means this should also take the deepest root level and not the hardcoded one.

Also, the `getState` method of setting the state is a great idea, but in my case not really usable because im nesting my states as well. See my example here:
```js
{
    MyComponent: {
        START() {},
        FirstState: {
            START() {},
            Confirmation: {
                Yes() {},
                No() {},
            }
        },
        SecondState: {
            START() {},
            Confirmation: {
                Yes() {},
                No() {},
            }
        }
    },
}
```
So if I want to go from `FirstState.Confirmation` to `SecondState`, the `getState()` method is not helpful because I need to get one state up. Thats why I was thinking about always taking the root as stating point for state navigation. Maybe a `getRootState` method should be also available to allow that functionality.

stammbach:
I think I've come up with a pretty good solution. I've added the `getRootState` method and implemented it onto `delegate`, `toStateIntent` & `toStatelessIntent`. Which means these methods will always use the current root state, e.g. None, `ComponentOne`, `ComponentOne.NestedComponent` without any internal states.

Also the ```this.toStateIntent(`${this.getState()}.TestState`, 'TestState');``` way is redundant because I could just use `this.toIntent('TestState');` 

I've modified the hello world skill example with some nested components and did some testing.

KaanKC:
> But then there is another problem, because I basically copied the logic from the sendComponentResponse method which means this should also take the deepest root level and not the hardcoded one.

`sendComponentResponse()` uses the `stateBeforeDelegate` property which is the value of `getState()` at the time of delegation. That should be the correct path at all times or am I missing something?

> So if I want to go from FirstState.Confirmation to SecondState, the getState() method is not helpful because I need to get one state up. Thats why I was thinking about always taking the root as stating point for state navigation. Maybe a getRootState method should be also available to allow that functionality.

Yeah, I see where the issue is. I mean technically that's something you could solve yourself inside the component but maybe providing a helper is the better solution.

> I think I've come up with a pretty good solution. I've added the getRootState method and implemented it onto delegate, toStateIntent, followUpState & toStatelessIntent. Which means these methods will always use the current root state, e.g. None, ComponentOne, ComponentOne.NestedComponent without any internal states.

Could you explain your solution in a little bit more detail, please? I feel like we could solve this without making changes to the existing routing functions.

I'd say we should simply add a function returning the full path to the currently active component's root state, which you can then use to call the routing methods. So in your example, it would look like this:

```js
// inside the `FirstState.Confirmation` state and you want to route to `SecondState`:
this.followUpState(`${this.getActiveComponentsRootStatePath()}.SecondState`);
```

Maybe with a different name for the function since it's pretty long.

stammbach:
There is no change with `sendComponentResponse`, because it always returns to the state the `delegate` was called from.

With my changes applied, `toStateIntent`, `followUpState` & `toStatelessIntent` always start at the (active component) root. Here are some examples:

toStateIntent inside a component (which is the root) just adds the provided state:
`FirstComponent` -> `this.toStateIntent('FirstState', 'START');` -> `FirstComponent.FirstState`

toStateIntent with a component name inside throws an JovoError because you should not be able to use a component name as state:
`FirstComponent` -> `this.toStateIntent('SecondComponent', 'START');` -> JovoError

toStateIntent from any state starts at the root so I can easily get back to the start of the app/active component:
`FirstComponent.FirstState.ConfirmationState` ... -> `this.toStateIntent('SecondState', 'START');` -> `FirstComponent.SecondState`

toStateIntent inside a nested component will then use the active component as root
`FirstComponent.SecondComponent` -> `this.toStateIntent('SomeState', 'START');` -> `FirstComponent.SecondComponent.FirstState`

there are no changes to the component usage which means wherever `this.delegate` gets called the component response will be send:
`FirstComponent.SomeState` -> `this.delegate('SecondComponent', { onCom..: 'CALLBACK' });` -> `FirstComponent.SecondComponent` -> `this.sendComponentResponse();` -> `FirstComponent.SomeState@CALLBACK`

I think this is much more seamless and easier to remember. It will always start from the root. So inside my app it always start from the "top" of my app. For a component it always starts at the "top" of the component. Always relying on string literals are not very user friendly, also because I did not expect it to leave my component in the first place.

stammbach:
I've published [my test app](https://github.com/stammbach/jovo-hello-nested-components), hope this helps with understanding what's going on. @KaanKC

jovo-framework/src/middleware/Handler.ts
```diff
@@ -334,9 +334,51 @@ export class Handler implements Plugin {
...
+            return this.internalToStateIntent(state, intent);
+        };
+
+        /**
+         * Jumps to state intent without any validation
+         * @private
+         * @param {string} state name of state
+         * @param {string} intent name of intent
+         */
+        Jovo.prototype.internalToStateIntent = async function (
```
KaanKC: Is it really necessary to have both `internalToStateIntent` and `toStateIntent`. Can't we just merge them?

jovo-framework/src/middleware/Handler.ts
```diff
@@ -334,9 +334,51 @@ export class Handler implements Plugin {
         Jovo.prototype.toStateIntent = async function (
             state: string | undefined,
             intent: string,
+        ): Promise<any> {
+            // Check for Component Names inside State
+            if (typeof state !== 'undefined' && this.$components) {
```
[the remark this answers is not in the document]
stammbach: Yes moving the 

> instead of using `if (typeof state !== 'undefined' && this.$components)` use ` if (state && this.$components)` please.
> 
> Keeps it consistent with the rest of the codebase.

I had that prior to the typeof stuff, but strangely I got some tslint errors with that, but I will try to change it back to simpler type checking. 

> I think we should move that part into its own function. That way we reduce some redundancy.

Yes, good point!

> My first idea as for the name of it would be `isStateSameAsIntializedComponentName()`. But since that's fairly long, I'm open for other alternatives.

There's nothing wrong with a long name, as **long** (pardon the pun) it's easier to understand. `checkStateForInitializedComponentName` would also fit in my option, also if you like it shorter we might just strip the "Initialized".

jovo-framework/src/middleware/Handler.ts
stammbach: We could, the only difference is the added validation on the new `toStateIntent`, `internalToStateIntent` is just the old method. But then there must be a way to toggle the validation using a parameter (which I don't like, because we should enforce the validation and not make it easy to disable) or a instance variable similar to `this.triggeredToIntent`

stammbach: I've added the `this.triggerStateValidation` property for now to determine if the state should be validated. But I can change it to whatever you prefer naming-wise.

stammbach:
@KaanKC I did some refinement to the codebase, I kept two old discussions unresolved because of naming-reasons. Waiting for your feedback on that. Otherwise everything should be fine now.

jovo-framework/src/middleware/Handler.ts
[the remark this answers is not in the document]
KaanKC: I like the name. Let's go with that.

jovo-framework/src/middleware/Handler.ts
KaanKC: I'd actually prefer using a parameter where we set the default value to `true`. It's more straightforward and is less prone to errors.

For example, with the current implementation, the validation wouldn't be run if we have the following scenario:

`this.delegate()` -> sets `triggerStateValidation` to false. -> in the same request&response, the component uses `this.toStateIntent()` -> since `triggerStateValidation` was set to false, the validation won't be triggered.

@aswetlow what's your take on it?

stammbach: Oh, didn't catch that, im sorry! Fixed by resetting the `triggerStateValidation` back to true before returning. Personally I'm not a huge fan of changing this to a parameter, because anyone who does't pay much attention might messing up the whole flow by unintentionally leaving/entering a component. The validation flag/toggle should remain internal use only, at least in my opinion.

aswetlow: Hey!
I think I would prefer Kaan's approach with the parameter. It makes the code a little more manageable and clearer to use (no additional variable in the Jovo class)

Great PR btw!

jovo-framework/src/middleware/Handler.ts
```diff
@@ -385,11 +405,69 @@ export class Handler implements Plugin {
...
+
+            this.triggerStateValidation = false;
             return this.toStateIntent(state, 'START');
         };
 
+        /**
+         * Checks if the given state contains the name of a initialized component.
+         * @throws {JovoError}
+         */
+        Jovo.prototype.checkStateForInitializedComponentName = function (state: string | undefined): void {
```
aswetlow: How about adding  `checkStateForInitializedComponentName` to the Handler class as a private method?

jovo-framework/src/middleware/Handler.ts
stammbach: Replaced it with the `validate` param.

jovo-framework/src/middleware/Handler.ts
stammbach: Moved it into the Handler class similar to the handle* methods, hope that's what you meant. Otherwise I'm not sure how to do "private" methods, I haven't had many visits to the TS-Camp yet.

aswetlow: It was in the right place. Can you put it back into the Jovo class? :)
Sorry, my mistake.

stammbach: Moved it back to the Jovo class. Hope everything is alright then. Really looking forward to a release with these changes :)

PR #627, opened 2019-11-28
Access reminder events with ON_EVENT

gshenar:
Proposed changes
Right now you cant access reminder events with ON_EVENT and Jovo errors out on the request since the type is not handled.

Maybe we want this with a new event like ON_REMINDER_EVENT? That may require more changes and updates to documentation though and I'm not sure it's needed.

https://developer.amazon.com/docs/smapi/alexa-reminders-api-reference.html#reminderstarted

## Types of changes
- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X ] My code follows the code style of this project


aswetlow:
Hey Gal! Thank you for the PR.

I think I would prefer adding a new type. To keep it simple, an additional condition in the SkillEvent class and ON_REMINDER_EVENT enum in `alexa-enums.ts` would be enough.

gshenar:
Sure - that makes sense

aswetlow:
As discussed, I close this one. 

Added in 4ed2f4521a603ad3b62d09bbb61cee6e5614e7b6

PR #701, opened 2020-03-28
🐛Fixes error when stopping the session

omenocal:
## Proposed changes
When trying to end a conversation with Autopilot, the core AutopilotResponse.ts file will invoke the hasSessionEnded(), which then calls (with no arguments) to the method isTell(), which expects 1 argument, speechText: string.

Since the method hasSessionEnded() didn't send any arguments, the logic will eventually crash with the following message:

Screen Shot 2020-03-28 at 4 37 17 PM

To end a conversation with autopilot, the actions array simply has to miss this item:
```
{
listen: true
}
```

For more information about listen, check this link https://www.twilio.com/docs/autopilot/actions/listen

## Types of changes

## Checklist

### DIFF
--- a/jovo-platforms/jovo-platform-twilioautopilot/src/core/AutopilotResponse.ts
+++ b/jovo-platforms/jovo-platform-twilioautopilot/src/core/AutopilotResponse.ts
@@ -107,27 +107,14 @@ export class AutopilotResponse implements JovoResponse {
     return this.setSessionAttributes(sessionData);
   }
 
-  isTell(speechText?: string | string[]): boolean {
+  isTell(): boolean {
     const hasListenAction = this.actions.some((action) => {
       return action.listen;
     });
     // is ask()!
     if (hasListenAction) return false;
 
-    const sayAction = this.actions.find((action) => {
-      return action.say;
-    });
-
-    // no speech output in response
-    if (!sayAction) return false;
-
-    if (Array.isArray(speechText)) {
-      for (const speech of speechText) {
-        if (AutopilotSpeechBuilder.toSSML(speech) === sayAction.say) return true;
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


jovo-platforms/jovo-platform-twilioautopilot/src/core/AutopilotResponse.ts
```diff
@@ -107,27 +107,14 @@ export class AutopilotResponse implements JovoResponse {
...
 
-  isTell(speechText?: string | string[]): boolean {
+  isTell(): boolean {
     const hasListenAction = this.actions.some((action) => {
       return action.listen;
     });
     // is ask()!
     if (hasListenAction) return false;
 
-    const sayAction = this.actions.find((action) => {
```
KaanKC: We actually need that deleted part for the TestSuite, 
e.g. `expect(response.isTell('Hello World!')).toBe(true);`

But, we should check whether `speechText` is defined:

```ts
isTell(speechText?: string | string[]): boolean {
  const hasListenAction = this.actions.some((action) => {
    return action.listen;
  });
  // is ask()!
  if (hasListenAction) return false;

  const sayAction = this.actions.find((action) => {
    return action.say;
  });

  // no speech output in response
  if (!sayAction) return false;

  if (speechText) {
    if (Array.isArray(speechText)) {
      for (const speech of speechText) {
        if (AutopilotSpeechBuilder.toSSML(speech) === sayAction.say) return true;
      }
    }

    return AutopilotSpeechBuilder.toSSML(speechText as string) === sayAction.say;
  }

  return true;
}
```

That should fix the error that came up.

aswetlow:
Fixed by @KaanKC in https://github.com/jovotech/jovo-framework/pull/707 (after consultation with @omenocal )

PR #713, opened 2020-04-07
🐛 Added log event message cast to string

RokasVaitkevicius:
## Proposed changes

Mongo database was throwing errors and one of the parameters in error object was `code` which was of type `number`. Because of that function `addConsoleAppender` in `Log.js` file was throwing this exception: 

```
TypeError: msg.split is not a function
      at Object.write (/../node_modules/jovo-core/dist/src/Log.js:135:45)
      ...
```

It was happening, because Number doesn't have `split` function, so I made sure that log event message is always of type `string`, by casting it.

After casting, I again was able to see exceptions thrown by my code,  instead of the one thrown by `addConsoleAppender` function.

## Types of changes

## Checklist


aswetlow:
Thanks.

PR #812, opened 2020-09-08
:bug: fix endreason for "stop" not set

freiSMS:
## Proposed changes
Stop is not an Alexa End Reason. Therefore it is not set by calling "getEndReason()". Now there is an additional check for intents which are closing the session but do not belong to the group above.  -> Mapped to "STOP" metric in gAnalytics. 

## Types of changes

## Checklist


freiSMS:
added another fix for setting custom dimensions @aswetlow

PR #875, opened 2020-12-08
persist user.context within unit tests :adhesive_bandage: 

fbublitz:
## Proposed changes

A call to conversation.send(request) in turn calls saveUserData(), which overwrites the userdb.json with the content of conversation.$user.$data.
If more than one request - response cycle is executed within one unit test, $user.$context is lost in between requests. This makes it impossible to test app behavior that is dependent on $context contents, since the conversation object doesn't allow for setting $context content before request execution either. 
Changing saveUserData() so that it just merges the contents of conversation.$user.$data into the userdb.json, allows for keeping accumulated $context. 
This solution is much simpler and safer than allowing for "hand" manipulating $context.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] My code follows the code style of this project
- [ x] All new and existing tests passed


ngocht:
The failing test seems to me an issue with Jest Timeout set to 550ms and a slow Windows VM. How to proceed ?
@aswetlow

jankoenig:
Thank you @fbublitz and @ngocht 🚀 
We'll take a look!

aswetlow:
Awesome! Thank you! I restarted the test and it worked.

PR #905, opened 2021-02-25
feat(GoogleAnalytics): support multiple experiments

theBenForce:
## Proposed changes
Allow users to track multiple Google Optimize experiments at once.

## Types of changes

## Checklist

### DIFF
--- a/jovo-integrations/jovo-analytics-googleanalytics/src/GoogleAnalytics.ts
+++ b/jovo-integrations/jovo-analytics-googleanalytics/src/GoogleAnalytics.ts
@@ -343,7 +343,7 @@ export class GoogleAnalytics implements Analytics {
     jovo.$googleAnalytics = {
       $data: {},
       $parameters: {},
-      $experiments: {} as Record<string, string | number>,
+      $experiments: {},
       sendEvent: (params: Event) => {
         this.visitor!.event(params, (err: any) => {
           if (err) {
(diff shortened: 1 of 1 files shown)


theBenForce:
@jankoenig I'd like to create some unit tests for `jovo-analytics-googleanalytics`. Are there any other packages that provide good examples to work off of?

Thanks!

aswetlow:
Hey @theBenForce 
Thanks for the PR.

Do you have something in mind you want to test? It always depends on the testcase.

Here's an example for testing full platform specific (request/responses)
https://github.com/jovotech/jovo-framework/blob/master/jovo-platforms/jovo-platform-alexa/test/FrameworkBase.test.ts

Here's an example mocking jovo methods
https://github.com/jovotech/jovo-framework/blob/a7d84081ec0858926459a483ff5986e12d35e5a3/jovo-core/test/Jovo.test.ts

You could also build simple jovo objects to test methods in GoogleAnalytics.ts. Something like this

```javascript
test('test getPageName', async (done) => {
  const googleAnalytics = new GoogleAnalytics();

  const jovo = ({
    $user: {
      $context: {},
      $data: {},
      $metaData: {},
    },
  } as unknown) as Jovo;

  expect(googleAnalytics.getPageName(jovo)).toBe('foobar');
});

``` 

Testing jovo plugins can be quite a hassle. This is also a big issue we want to improve in v4.

theBenForce:
@aswetlow thanks for the info. I wanted to add some tests to validate the format of experiments being added. Not sure if it's worth the hassle since it's such simple logic.

PR #910, opened 2021-03-09
Google Analytics: Custom Metrics and Dimensions supporting names + validation

freiSMS:
## Proposed changes
- Allow developers to define the index of google Analytics metrics and dimensions set by system (see here https://community.jovo.tech/t/google-analytics-custom-dimension-metrics-improvements/1510?u=andre)
  - fixes problem of overwriting developers custom definitions
  - readable code -> instead of setting "cd1" your are setting custom dimension 'UUID'
  - allow changing target custom definition without changing code logic
- Add "strict mode" via config -> will check if all needed system definitions are defined + are not mapped to same index
- make $googleAnalytics instance of class 
  -  -> gives access to features of used universal-analytics
  -  -> enhance class with new method will no longer need second declaration on $googleAnalytics object and third definition on interface

## Checklist


aswetlow:
Thank you so much, André

PR #927, opened 2021-05-02
 Google Analytics Allow overriding documentTitle/documentPath/documentHostName

kouz75:
## Proposed changes

Adding customParameters at the end of the return object will allow us to override  documentTitle/documentPath/documentHostName.

Add following command will override default page title by "MyIntentName"
this.$googleAnalytics.setParameter('documentTitle','MyIntentName');

## Types of changes

## Checklist

### DIFF
--- a/jovo-integrations/jovo-analytics-googleanalytics/src/GoogleAnalytics.ts
+++ b/jovo-integrations/jovo-analytics-googleanalytics/src/GoogleAnalytics.ts
@@ -336,10 +336,10 @@ export class GoogleAnalytics implements Analytics {
     const customParameters = jovo.$googleAnalytics.$parameters;
 
     return {
-      ...customParameters,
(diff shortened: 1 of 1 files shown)


aswetlow:
Thanks

freiSMS:
Nice idea @kouz75 . We also overwrite some Google Analytics methods. We use class inheritance. Maybe this could be useful for you in cases where you want to always overwrite them. The protected Google Analytics methods like getPageName(), getUserId() etc. are intended to cover cases where developers want to have their own behavior. For example you could use your own userIds (like bellow) or do some stuff before sending data via the track method.
``` javascript
export class CustomGoogleAnalyticsAlexa extends GoogleAnalyticsAlexa {
    protected getUserId(jovo: Jovo): string {
      return Custom.userIdFromDatabase()
}
```

PR #929, opened 2021-05-06
✨ added sessionId to ContextPrevObject

JayDotDot:
## Proposed changes
This PR adds the sessionId of the request to the ContextPrevObject. A lot of times, the only past interactions of interest are the ones of the same session or a specific session. This adds this session context to the prev array. It also makes it easier to skip for the session irrelevant interactions. One Example would be Alexa Skill Events that happen in the middle of the session and therefore are added to the prev array in middle of all the session requests with no easy way to see, that this interaction was not part of the session. 

I guess this would require an update to the documentation for the user data (https://www.jovo.tech/docs/data/user). I could also add the documentation, if this feature is to be added.

## Types of changes

## Checklist

### DIFF
--- a/docs/basic-concepts/data/user.md
+++ b/docs/basic-concepts/data/user.md
@@ -52,6 +52,7 @@ user: {
             request: {
                 intent: true,
                 state: true,
+                sessionId: true,
                 inputs: true,
                 timestamp: true,
             },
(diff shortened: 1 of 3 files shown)


aswetlow:
Hey @JayDotDot 

Thanks for the PR. I talked with @dominik-meissner about that topic a while ago. Happy, you made it configurable 😄 

It would be nice if you would update the docs as well.

JayDotDot:
Hey @aswetlow, 
I just realized that all these docs are also part of the repo, makes sense. I added the documentation. While doing that I saw, that the other properties for the ContextPrevObject have their own getter methods like `this.$user.getPrevIntent(i)` that are however deprecated. Not sure, if adding a deprecated function for consistency, or just omitting this is the way to go.

aswetlow:
Thank you. I will publish it tomorrow.

PR #940, opened 2021-06-23
:sparkles: Implement passing of OutputTemplate-properties directly to `$send` via the options

m-ripper:
## Proposed changes
It is now possible to pass properties of `OutputTemplate` to `$send`, e.g.  `$send(SomeOutput, { message: 'Overwritten message' })` will result in an `OutputTemplate` with `message` set to 'Overwritten message'.

## Types of changes

## Checklist

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -149,13 +149,43 @@ build() {
 
 ### Output Options
 
-As a convention, an output template should only be responsible for organizing the output, not collecting any data. To achieve this, the handler should first collect all necessary information and the pass it to the output class as `options`:
+As a convention, an output template should only be responsible for organizing the output, not collecting any data. To achieve this, the handler should first collect all necessary information and then pass it to the output class as `options`:
 
 ```typescript
 return this.$send(YourOutput, { /* options */ });
 ```
 
-These options can then be referenced inside the output class using `this.options`.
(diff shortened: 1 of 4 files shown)


aswetlow (approved):

jankoenig:
If I understand it correctly, it only works for one level right now: 'message', 'reprompt', 'listen', 'quickReplies', 'card', 'carousel', 'platforms'.

What if an output template has the following:

```js
{
  // ...
  platforms: {
    GoogleAssistant: {
      // ..
    }
  }
}
```

And in the `$send` method we pass the following:

```js
this.$send(SomeOutput, { platforms: { Alexa: { /* platform specific stuff */ } } }
```

Would the result then be the following (removing the Google Assistant property):

```js
```js
{
  // ...
  platforms: {
    Alexa: {
      // ..
    }
  }
}
```

Could this be solved by offering a `merge` instead of a `replacement` instead?

m-ripper:
Yes, it would probably make sense to merge `platforms` and it's contents instead of replacing the value. 
I will adjust the implementation.

jankoenig:
Great, thanks @m-ripper, I just added docs. Could you take a quick look if I missed anything? If not, we're good to go. Thanks!

PR #943, opened 2021-06-28
✨ Implement global components

m-ripper:
## Proposed changes
- The `@Global`-decorator is now dynamic and can be applied to methods and classes
- Components can be defined as global by passing `global: true` to the options or by using the `@Global`-decorator
- All handlers of a global component are global as well
- Global components do not get added to the state-stack, with the exception being $delegate, where the global component is added to the state-stack but just until the delegation was resolved.

## Types of changes

## Checklist


aswetlow (approved):

jankoenig (approved):
Thanks! I just added the documentation. Good to go from my side

PR #945, opened 2021-06-30
Rasa NLU lautmaler edition

fbublitz:
## Proposed changes
Hello i have extended the v4 Rasa NLU plugin towards full functionality
I can also write Documentation and unit tests if you instruct me how.

- added a constructor with a super() call because a user config object was not handed down
- extended rasa intent and entity interface definitions
- extended the NluData interface to RasaNluData interface 
- intent classification confidence is mapped into the nlu result now as well
- added entity mapping from rasa response, please check if the mapping of rasa entity role is ok
- added configurable mapping of rasa intent ranking into the nlu result
    -  switch on/off constrain amount mapped classification alternatives

## Types of changes

## Checklist

- [?] My code follows the code style of this project


integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -49,4 +96,14 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
   ): Promise<AxiosResponse<RasaResponse>> {
     return axios.post(`${this.config.serverUrl}${this.config.serverPath}`, { text }, config);
   }
+
+  private mapAlternativeIntents(altIntents: RasaIntent[]) {
+    // remove first element, because its the classified intent
+    altIntents.splice(0, 1);
```
m-ripper, line 102: If the response from Rasa is persisted in the future (which can and can not happen), this could cause some side effects due to mutating the parameter `altIntents`. In that case, it would probably be better to simply `slice` before filtering.

package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "version": "4.0.0",
   "scripts": {
-    "postinstall": "npm run bootstrap",
+    "postinstall": "npm run   bootstrap",
```
m-ripper, line 6: This was probably just a mistake, but it should be just a single whitespace.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -49,4 +96,14 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
   ): Promise<AxiosResponse<RasaResponse>> {
     return axios.post(`${this.config.serverUrl}${this.config.serverPath}`, { text }, config);
   }
+
+  private mapAlternativeIntents(altIntents: RasaIntent[]) {
```
m-ripper, line 100: A return-type is required to satisfy eslint.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -3,19 +3,30 @@ import {
...
   Jovo,
-  JovoError,
   NluData,
   NluPlugin,
   PluginConfig,
 } from '@jovotech/framework';
-import { RasaResponse } from './interfaces';
+import { RasaEntity, RasaIntent, RasaResponse } from './interfaces';
 
 export interface RasaNluConfig extends PluginConfig {
```
m-ripper, line 15: In v4, the Config-type of a plugin is supported to exactly reflect the config as it should be, meaning if the integration needs a `serverUrl` like here for example, then it should not be optional in the `RasaNluConfig`.
Only properties that are really optional and not required for the integration to work are now supposed to be optional.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -49,4 +96,14 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
   ): Promise<AxiosResponse<RasaResponse>> {
     return axios.post(`${this.config.serverUrl}${this.config.serverPath}`, { text }, config);
   }
+
+  private mapAlternativeIntents(altIntents: RasaIntent[]) {
+    // remove first element, because its the classified intent
+    altIntents.splice(0, 1);
+    const cutoff = this.config.alternativeIntents?.confidenceCutoff ?? 1.0;
```
m-ripper, line 103: Instead of using a default value here, it is recommended to use `getDefaultConfig` and use the config-value directly.
In this example it would mean, that `RasaConfig.alternativeIntents` and `RasaConfig.alternativeIntents.confidenceCutoff` become required and `getDefaultConfig` returns `alternativeIntents.confidenceCutoff` as well.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -49,4 +96,14 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
...
+
+    return altIntents
+      .filter((a) => a.confidence > cutoff)
+      .slice(0, this.config.alternativeIntents?.maxAlternatives ?? 7);
```
m-ripper, line 107: Same as the previous comment.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -3,19 +3,30 @@ import {
...
-  serverUrl: string;
-  serverPath: string;
+  serverUrl?: string;
+  serverPath?: string;
+  //activate alternative intent classifications in $nlu
+  //and constrain the number of alternatives by number and/or an confidence cutoff
+  alternativeIntents?: { maxAlternatives?: number; confidenceCutoff?: number };
+}
+
+interface RasaNluData extends NluData {
```
m-ripper, line 23: Due to the interface not being exported, this is not used anywhere outside of `RasaNlu`. 
It should either get exported or get removed. The main issue is, that the type of `Jovo.$nlu` will always be `NluData` without casting.

m-ripper (changes requested):
Hey there, first of all, thanks for the pull request and sorry that it took so long for us to review.

A few changes have to be made but none of them are conceptually but mostly code-style and such. 

If you have more questions, feel free to ask.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -29,14 +40,50 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
     };
   }
 
-  async process(handleRequest: HandleRequest, jovo: Jovo): Promise<NluData | undefined> {
+  constructor(config?: RasaNluConfig) {
```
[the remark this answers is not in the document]
fbublitz: Removed. I think at that time the config merge didnt work thats why i had my own constructor

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -29,14 +40,50 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
...
     const text = jovo.$request.getRawText();
     if (!text) return;
     try {
       const rasaResponse = await this.sendTextToRasaServer(text || '');
-      return rasaResponse?.data?.intent?.name
-        ? { intent: { name: rasaResponse.data.intent.name } }
-        : undefined;
+      if (rasaResponse.data.intent.name) {
+        const ents: EntityMap = {};
+        rasaResponse.data.entities.map((entity: RasaEntity) => {
```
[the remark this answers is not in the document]
fbublitz: Changed to reduce

package.json
fbublitz: adapted

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: added return type

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: all config elements are now mandatory. alternative intent mapping should have been optional, depending on if the config option is given or not. but i am totally fine with having it as a fixed part. The change led to a considerable code restructure.

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: adapted in consequence of the above

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: adapted in consequence of the above

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: decided to export it. Plus in consequence of the other changes the return type of _process_ needs to be _RasaNluData_ now.

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: renamed method argument and using slice now

fbublitz:
upstream v4dev was fetched and merged as well

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -3,17 +3,30 @@ import {
...
   NluPlugin,
   PluginConfig,
 } from '@jovotech/framework';
-import { RasaResponse } from './interfaces';
+import { RasaEntity, RasaIntent, RasaResponse } from './interfaces';
 
 export interface RasaNluConfig extends PluginConfig {
-  serverUrl: string;
-  serverPath: string;
+  serverUrl?: string;
```
m-ripper, line 16: `serverUrl` and `serverPath` should not be optional here due to being required to call the Rasa server.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -24,20 +37,30 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
...
+      if (rasaResponse.data.intent.name) {
+        return {
+          intent: {
+            name: rasaResponse.data.intent.name,
+            confidence: rasaResponse.data.intent.confidence,
+          },
+          alternativeIntents: this.mapAlternativeIntents(rasaResponse.data.intent_ranking),
+          entities: rasaResponse.data.entities.reduce(RasaNlu.mapEntities, {}),
+        };
+      } else {
```
m-ripper, line 58: Code can be shortened by removing `else`. If the intent-name exists, `return` is called anyways and therefore the `else`-block can be omitted.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -24,20 +37,30 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
...
+      alternativeIntents: { maxAlternatives: 15, confidenceCutoff: 0.0 },
     };
   }
 
-  async process(handleRequest: HandleRequest, jovo: Jovo): Promise<NluData | undefined> {
+  async process(handleRequest: HandleRequest, jovo: Jovo): Promise<RasaNluData | undefined> {
     const text = jovo.$request.getRawText();
     if (!text) return;
     try {
       const rasaResponse = await this.sendTextToRasaServer(text || '');
```
m-ripper, line 48: Alternative value of `''` is not needed due to checking if the text exists before.

lerna.json
```diff
@@ -4,7 +4,8 @@
     "framework",
     "platforms/**",
     "integrations/**",
-    "examples/**"
+    "examples/**",
+    "../botify/chatbots/team/botify-team-bot/"
```
m-ripper, line 8: This was accidentally added, please remove for the PR.

integrations/nlu-rasa/src/RasaNlu.ts
```diff
@@ -48,4 +71,30 @@ export class RasaNlu extends NluPlugin<RasaNluConfig> {
...
+  private mapAlternativeIntents(allIntents: RasaIntent[]): RasaIntent[] {
+    // remove first element, because its the classified intent
+    const alternativeIntents = allIntents.slice(1);
+
+    return alternativeIntents
+      .filter((a) => a.confidence > this.config.alternativeIntents.confidenceCutoff)
+      .slice(0, this.config.alternativeIntents.maxAlternatives);
+  }
+
+  private static mapEntities(entityMap: EntityMap, rasaEntity: RasaEntity): EntityMap {
```
m-ripper, line 84: Not sure if this has to be static, theoretically an instance-method should suffice.

m-ripper (changes requested):
Thanks again for the changes, we're almost there! A few changes still have to be made.

If you have any more questions, feel free to ask.

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: Done. thought i did that already

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: Done

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: i was wondering if this check is necessary or sensible too. Removed

lerna.json
fbublitz: Done

integrations/nlu-rasa/src/RasaNlu.ts
fbublitz: i thought it is good practice to make methods static that do not access any instance variables. What do you think?

m-ripper: I've read that too somewhere but I am not a huge fan of that personally. Being able to access the method via `this` seems cleaner to me and there should not be a noticeable difference in terms of performance either.

 I guess going with the instance method would be better just for the sake of fitting the code style of the other packages.

fbublitz: Ok that's fine with me. Done.

jankoenig:
Thank you @fbublitz and @m-ripper! 🎉 

I added documentation to the `pr/945` branch: https://github.com/jovotech/jovo-framework/blob/pr/945/integrations/nlu-rasa/README.md

Ready to merge from my side ✅

jankoenig (approved):

m-ripper:
@fbublitz can you please merge the `origin/pr/945`-branch into the branch of this PR? Then we should be ready to merge!

fbublitz:
Ok, pr/945 branch is merged. thanks a lot!

jankoenig:
One thing that I mentioned during the office hours: With the #981, we removed the `$nlu` property in favor of the `$input` property.

So the alternative intents feature would be accessed like this: `this.$input.nlu.alternativeIntents`. I would suggest we do this for now and, if there's a need, re-add a convenience `$nlu` property at a later point?

m-ripper:
You can also enable commits by maintainers for this PR and I will be happy to take over this task if you like.

fbublitz:
the access to alternativeIntents via _input.nlu_ is totally fine for me.

m-ripper (approved):

PR #946, opened 2021-06-30
✨♻️ Implement Jovo History functionality, recycle persistence layer

aswetlow:
## Persisting in general
In v4 the persistence layer will be more flexible compared to v3.
The type of data that is stored is configurable. The developer of a Jovo app
defines the level of data that is stored in the database configuration. 

### Default configuration:

`app.dev.ts`
```typescript
app.use(
  new FileDb({
    pathToFile: './db/db.json',
  }),
);
```

This will create the following database object in `db/db.json`

```json
[
  {
    "id": "amzn1.account.AM3B00000000000000000000000",
    "user": {
      "data": {
       }
    },
    "createdAt": "2021-06-30T05:50:04.034Z",
    "updatedAt": "2021-06-30T05:50:04.095Z"
  }
]
```

**New in v4**: `createdAt` and `updatedAt` are stored by default. This helps debugging users in large datasets.

### Custom Configuration

`app.dev.ts`
```typescript
app.use(
  new FileDb({
    pathToFile: './db/db.json',
    storedElements: {
      user: true, // this.$data.user, optional, shortcut for {enabled: true}
      session: true, // this.$session, optional
      history: {
          enabled: true,
          size: 3, // size of this.$history array
          output: true, // this.$output, optional
          state: true, // this.$state, optional
          nlu: true, // this.$nlu, optional
          entities: false, // this.$entities, optional
          asr: false, // this.$asr, optional
          request: false, // this.$request, optional
          response: false, // this.$response, optional
          myCustomData: (jovo: Jovo) => {
              return `my custom data for user: ${jovo.$user.id}`
          }
      },
    }, 
      
  }),
);
```
`db/db.json`
```json
[
  {
    "id": "amzn1.account.AM3B00000000000000000000000",
    "user": {
      "data": {}
    },
    "session": {
      "id": "amzn1.echo-api.session.0000000-0000-0000-0000-00000000000",
      "data": {},
      "state": [
        {
          "componentPath": "LoveHatePizzaComponent"
        }
      ],
      "createdAt": "2021-06-30T06:45:40.444Z",
      "updatedAt": "2021-06-30T06:47:44.250Z"
    },
    "history": [
      {
        "output": {
          "message": "Yes! I love pizza, too."
        },
        "nlu": {
          "intent": {
            "name": "AMAZON.YesIntent"
          },
          "entities": {}
        },
        "state": [
          {
            "componentPath": "LoveHatePizzaComponent"
          }
        ],
        "myCustomData": "my custom data for user: amzn1.account.AM3B00000000000000000000000"
      },
      {
        "output": {
          "message": "Hello World! Do you like pizza?",
          "listen": true
        },
        "nlu": {},
        "state": [
          {
            "componentPath": "LoveHatePizzaComponent"
          }
        ],
        "myCustomData": "my custom data for user: amzn1.account.AM3B00000000000000000000000"
      }
    ],
    "createdAt": "2021-06-30T06:45:40.444Z",
    "updatedAt": "2021-06-30T06:47:44.253Z"
  }
]
```

The history array is sorted by time (DESC), means the latest history item is on position `this.$history[0]`


integrations/db-filedb/src/FileDb.ts
```diff
@@ -16,25 +8,33 @@ export interface FileDbConfig extends DbPluginConfig {
...
-}
+export class FileDb extends DbPlugin<FileDbConfig> {
+  constructor(config?: FileDbConfig) {
+    super(config);
+  }
 
-export class FileDb extends Plugin<FileDbConfig> {
   getDefaultConfig(): FileDbConfig {
     return {
       pathToFile: './db/db.json',
```
jankoenig, line 20: Was the v3 default `../db/db.json`? If I see it correctly, this would result in the `db` being stored in `src/db` instead of `db`

aswetlow: Yes, `../db/db.json` would be probably clearer since `app.dev.ts` is in `/src`. 
Another option could be:
```
app.use(
  new FileDb({
    pathToFile: __dirname + '../db/db.json',
    })
);
```

framework/src/Jovo.ts
```diff
@@ -408,6 +421,37 @@ export abstract class Jovo<
     return this.$session.isNew;
   }
 
+  getPersistableData(): JovoPersistableData {
+    return JSON.parse(
```
m-ripper, line 425: I am assuming JSON.stringify followed by JSON.parse is used to get rid of references and get a deep copy?
Or is it also used to get rid of functions and class-instances?

In case it's just used for getting a deep copy we could either use `lodash.clonedeep`, which is a dependency already, or make sure that `getPersistableData` returns a copy already and clone the `history` via `slice` although this would not deep-clone the items of `history`. But for that issue, a mapper could be used.

integrations/db-filedb/src/FileDb.ts
```diff
@@ -16,26 +8,16 @@ export interface FileDbConfig extends DbPluginConfig {
...
+export class FileDb extends DbPlugin<FileDbConfig> {
+  constructor(config?: FileDbConfig) {
+    super(config);
+    console.log(this.config);
+  }
 
-export class FileDb extends Plugin<FileDbConfig> {
   getDefaultConfig(): FileDbConfig {
     return {
+      // ...super.getDefaultConfig(),
```
m-ripper, line 19: Is this intentionally disabled? Wouldn't this cause the config to not contain the default-config of `DbPlugin` because it never gets called?

framework/src/Jovo.ts
aswetlow: Thanks, removed stringify->parse. It wasn't necessary here.

integrations/db-filedb/src/FileDb.ts
aswetlow: This was a WIP commit. The double slash is removed

m-ripper (approved):

jankoenig (approved):
Approved! Looks great

m-ripper (approved):
Look

PR #962, opened 2021-07-19
🚚 Rename platform resources

rubenaeg:
## Proposed changes
This PR renames platform resources according to the rule to append the plugin type to it's constructor, e.g. Alexa -> AlexaPlatform, FacebookMessenger -> FacebookMessengerPlatform.

## Types of changes

## Checklist


docs/project-config.md
```diff
@@ -0,0 +1,279 @@
...
+
+You can also use this shorthand for nested folders:
+
+```js
+const project = new ProjectConfig({
+  // ...
+  plugins: [
+    new AlexaCli({
+      files: {
+        'path.to.file.json': 'Hello World!'
```
rubenaeg, line 139: This wouldn't work, as a dot is used to mark the file extension. I'll adjust this to use slashes instead.

docs/project-config.md
```diff
@@ -0,0 +1,279 @@
...
+});
+```
+
+By default, only the elements outside the `stages` are used. If a specific stage is active, all content from this stage is merged into the generic config.
+
+The active stage is determined in the following order (most active first):
+
+* The stage is added as flag, e.g. `jovov4 build --stage someStage`
+* The default stage is set with `defaultStage: 'someStage'`
+* The stage is set in the environment variables with `NODE_ENV=someStage`
```
rubenaeg, line 208: Currently, the implementation takes the `defaultStage` as the last option. The following order is correct:

1. Flag, e.g. `jovov4 build --stage someStage`
2. `process.env.JOVO_STAGE`
3. `process.env.NODE_ENV`
4. `defaultStage: someStage`

Is this something we should discuss further?

docs/project-config.md
```diff
@@ -0,0 +1,279 @@
...
+  // ...
+  hooks: {
+    'before.build': [
+      () => console.log('Starting the build process now');
+    ]
+  },
+});
+```
+
+However, hooks usually require a few more lines of code. We recommend placing each function in a separate file in a `hooks` folder and then referencing it in the `jovo.project.js` file ([you can find an example of this hook here](https://github.com/rubenaeg/fetch-language-model-test/blob/master/hooks/fetchLanguageModel.hook.js)):
```
rubenaeg, line 252: Would it be useful to host the example on a `jovotech/` repository?

rubenaeg (commented):
@jankoenig I've added a few comments on the documentation.

platforms/platform-core/src/index.ts
```diff
@@ -14,14 +14,14 @@ declare module '@jovotech/framework/dist/types/Extensible' {
 
 declare module '@jovotech/framework/dist/types/Jovo' {
   interface Jovo {
-    $corePlatformApp?: CorePlatformApp;
+    $corePlatform?: Core;
   }
 }
-registerPlatformSpecificJovoReference('$corePlatformApp', CorePlatformApp);
+registerPlatformSpecificJovoReference('$corePlatform', Core);
```
m-ripper, line 20: @rubenaeg
This should be `$core` instead of `$corePlatform` or did we discuss something about this that I've missed?
The same goes for the module augmentation above.

rubenaeg: You are right, as we discussed I pushed a change that adjusts the code accordingly.

docs/project-config.md
jankoenig: Ah yes, this makes sense! Thank you, I'll update

docs/project-config.md
jankoenig: I think we can keep it for now while we're still in Alpha, and then add example to the official CLI repo?

jankoenig (approved):
Thank you! I just updated the docs. Should be good to go now 👍🏻

docs/project-config.md
rubenaeg: Sounds good. I'll update the repo then to be public instead of private.

PR #963, opened 2021-07-19
✨  Update i18n

jankoenig:
This PR comes with newly added docs for the `i18n` feature in Jovo `v4`. 

The documentation already includes features that are not implemented yet. Before we can merge, the following features need to be added:

* i18n should have `returnObjects: true` as default value so that objects and arrays work in the same way as in Jovo v3
* If a key returns an array, the integration should randomly pick one of its items


jankoenig:
Thanks @aswetlow! I just updated the documentation

aswetlow (approved):

aswetlow:
Good to go, from my side.

PR #966, opened 2021-07-20
✨ Implement this.$device (wip)

aswetlow:
First implementation of the `this.$device` object. 

Snippets from our design document (and internal discussions)

## Device Capabilities in v3
```javascript
// Generic

this.hasScreenInterface() || this.$request.hasScreenInterface()
this.hasVideoInterface() || this.$request.hasVideoInterface()
this.hasAudioInterface() || this.$request.hasAudioInterface()
this.$request.getDeviceName()

// Platform-specific
this.$googleAction.hasScreenInterface()
this.$googleAction.hasWebBrowserInterface()
this.$alexaSkill.hasAPLInterface() || this.$alexaSkill.$request.hasAPLInterface()
```

## Motivation for v4

There should be a clear way to find out about a device's capabilities. All these helper methods should be grouped under a class, not the root `this` context → `this.$device`.

- The `interface` term was mainly inspired by Alexa, which has different interfaces like AudioPlayer, APL, etc... → does `capabilities` fit better?
- Could we find ways to determine both `input` (speech, touch, type/keyboard) and `output` (audio, video, visual) capabilities?
- Do we need platform specific methods like `this.$alexaSkill.hasAPLInterface()`
- Could there be a consistent way to write a device's capabilities, e.g. through hooks or plugins?

## $device in v4

The JovoDevice class is abstract and has to be implemented by every platform with a few generic and a couple of specific functions. The platform fills the `capabilities` array on initiation. `supports(capability: Capability[] | Capability)` to check whether a request supports a specific device type.

Example for Alexa

```typescript
private setCapabilitiesFromRequest(): void {
    const supportedInterfaces = this.jovo.$request.context?.System?.device?.supportedInterfaces;

    if (supportedInterfaces?.AudioPlayer) {
      this.addCapability('audio', 'long-form-audio');
    }

    if (supportedInterfaces?.['Alexa.Presentation.APL']) {
      this.addCapability('screen', 'Alexa.Apl');
    }
    ...
  }
```

## Usage (v0.1)

 (cross-platform)

```typescript

// in component/output

if (this.$device.supports(['screen'])) {
   // do stuff for devices with screens
}

```

(platform-specific)

```typescript

// in component/output

await this.$alexaSkill.$device.getDeviceAddress(); // not implemented yet

const deviceId = this.$alexaSkill.$device.id

```


framework/src/JovoDevice.ts
```diff
@@ -0,0 +1,47 @@
...
+> {
+  capabilities: CAPABILITY[] = [];
+
+  public constructor(readonly jovo: JOVO) {
+    this.setCapabilitiesFromRequest();
+  }
+
+  protected abstract setCapabilitiesFromRequest(): void;
+
+  // TODO: AND or OR ? currently OR
```
rubenaeg, line 29: I'm not sure what you mean here, but if you mean `AND` in the sense of the Intersection Type I think this wouldn't really work here.

rubenaeg (approved):

framework/src/JovoDevice.ts
m-ripper: I also think that the feature can be omitted because developers can simply use `supports` multiple times. That way developers have much more control as well. In complex situations, helper functions could be implemented.

PR #983, opened 2021-08-17
V3/bugfix/facebook messenger platform

brody-ly:
## Proposed changes

This PR addresses some issues I've faced when working on a Jovo v3 app that makes use of the Facebook messenger platform. I thought it would be good to list them here to raise awareness of these issues for when the Facebook messenger platform gets ported to v4. 

### Unpublished `sample-request-json` files
The `sample-request-json` files are not published, this would result in an error when anyone would attempt to write test when using the Facebook messenger platform. Example:

sample.test.ts
```
import { FacebookMessenger } from 'jovo-platform-facebookmessenger';

const facebookMessengerConfig = { ... }

for (const p of [new FacebookMessenger(facebookMessengerConfig) ]) {
    const testSuite = p.makeTestSuite();

    . . . 
   const launchRequest = await testSuite.requestBuilder.launch();
```

Calling `.launch()` would return the following error when you run `npm run test`

```
Cannot find module '../../../sample-request-json/v1/LaunchRequest.json' from 'node_modules/jovo-platform-facebookmessenger/dist/src/core/FacebookMessengerRequestBuilder.js'
```

### Postback buttons would always invoke the ON_TEXT intent
Anytime a user would interact with a postback button, [the user would always be directed to the ON_TEXT intent](https://github.com/brody-ly/jovo-framework/commit/5786e95d05cdb4dc00c1126289c7abec9c1cce05). I believe the correct behaviour should be to route the user to the invocation of the payload string in the postback button. 

Example:
`new PostbackButton("Settings", "SettingsIntent")` should take you the SettingsIntent not ON_TEXT intent

### Facebook messenger platform response
When implementing the Facebook platform, I realised that there was no response returning from the server. This meant that it wasn't possible for me to perform any testing as there was no response returning. 

What I ended up doing was to add in the middleware [`after.response`](https://github.com/brody-ly/jovo-framework/commit/5d5c33a79b2a7bf4bcc1cbad681ec336e05c3648#diff-abf3ecb6ebac910e2e7c6911e6a940b8cd8cadd8cbc7255b39397ef148675bc9R108) to return the response. 

While addressing this issue, I realised that not all helper methods of the Facebook class returned a response. As a fix, I've introduced [a new key in the `$output` object](https://github.com/brody-ly/jovo-framework/commit/10f0855c03fe86ee0dc25c452195a3f5eb2659e0#diff-2d007621a74a9883a19e3143c29799a41ace0f0ac2b49a76d1c92c954bf5ab7aR123-R128) that stores all responses in an invocation. This change introduces a new response for the Facebook messenger platform. An example of this would look something like this,

App.ts
```
app.setHandler({
  LAUNCH() {
    return this.toIntent("GreetingIntent");
  },

  async GreetingIntent() {
    try {
      await this.$messengerBot?.showAttachment({
        type: AttachmentType.Image,
        data:
          "https://via.placeholder.com/200x100",
      });
      await this.$messengerBot?.showButtonTemplate({
        text: `G'day ${this.$messengerBot.$session.$data.userProfile.first_name}, would you like to subscribe?`,
        buttons: [
          new PostbackButton(
            `Yes please!`,
            "SubscribeIntent"
          ),
          new PostbackButton("No thanks", "DenySubscribeIntent"),
          new PostbackButton("Maybe? Not sure", "ConvinceSubscribeIntent"),
        ],
      });
      return;
    } catch (e) {
      console.error(e.response?.data || e.message);
      return;
    }
  }
});
```
The output of a launch request would look like this,

```
{
   "message": [
      {
         "recipient": {
            "id": "xxxxxxxx"
         },
         "options": {
            "type": "image",
            "data": "https://via.placeholder.com/200x100"
         }
      },
      {
         "recipient": {
            "id": "xxxxxxxxx"
         },
         "message": {
            "attachment": {
               "payload": {
                  "text": `G'day Brody, would you like to subscribe?",
                  "buttons": [
                     {
                        "type": "postback",
                        "title": "Yes please! 😄",
                        "payload": "SubscribeIntent"
                     },
                     {
                        "type": "postback",
                        "title": "No thanks",
                        "payload": "DenySubscribeIntent"
                     },
                     {
                        "type": "postback",
                        "title": "Maybe? Not sure",
                        "payload": "ConvinceSubscribeIntent"
                     }
                  ],
                  "template_type": "button"
               },
               "type": "template"
            }
         }
      }
   ]
}
```
Receiving a response like this would make testing possible.

### Facebook messenger typing actions
The current implementation of messenger typing action works as intended, however I found it hard to write tests that would test the delay between typingOn and typingOff action.

Current implementation
App.ts
```
await this.$messengerBot?.showAction(SenderActionType.TypingOn);
await delay(2000);
await this.$messengerBot?.showAction(SenderActionType.TypingOff);
await this.$messengerBot?.showText({
   text: `Hello`,
});
```
With the above implementation, I feel like there is no way to test the delay.

My proposal on a fix would be to do implement something similar to the following,

Called somewhere in App.ts
```
await this.$messengerBot?.showTyping(2000);
```

Which would respond with the following result,

```
{
    messages: [
      {
        typingOn: {
          recipient: {
            id: 'xxxxxxx',
          },
          sender_action: 'typing_on',
        },
        typingOff: {
          recipient: {
            id: 'xxxxxxx',
          },
          sender_action: 'typing_off',
        },
        delayInMs: 2000,
      },
    ],
  };
```
The above response would make it possible to test how long of a pause occurred between typingOn and typingOff. I've implemented a new helper method that does this [here](https://github.com/brody-ly/jovo-framework/commit/b3b0be696aba905d544592a4257dda0adc3d51b3).

I know the Jovo team is busy with the Jovo v4 release, but again, I'd like to reiterate that I thought It would be good to create this PR to raise awareness of the issues I've faced. I would love to get any sort of feedback on the changes in the PR and whether these issues are addressed in Jovo v4. I've held off writing tests until I've received feedback on whether these changes are worth merging or not.  

### Persistent menu
Added a feature that allows the platform to configure [messenger's persistent menu](https://developers.facebook.com/docs/messenger-platform/send-messages/persistent-menu/). 

## Types of changes

## Checklist

- [ ] I have updated the documentation accordingly ( I will update this based on feedback)
- [ ] I have added tests to cover my changes ( I will update this based on feedback)

### DIFF
--- a/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
+++ b/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
@@ -105,6 +105,7 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
 
     app.middleware('platform.output')!.use(this.output.bind(this));
     app.middleware('response')!.use(this.response.bind(this));
+    app.middleware('after.response')!.use(this.afterResponse.bind(this));
 
     this.use(new FacebookMessengerCore());
 
@@ -288,6 +289,13 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
         Log.error(e.response?.data);
       }
     }
+  }
+
+  async afterResponse(handleRequest: HandleRequest) {
+    if (!handleRequest.jovo || handleRequest.jovo.constructor.name !== this.getAppType()) {
+      return Promise.resolve();
+    }
+    await handleRequest.host.setResponse(handleRequest.jovo.$response);
   }
 
   makeTestSuite(): TestSuite<FacebookMessengerRequestBuilder, FacebookMessengerResponseBuilder> {
--- a/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
+++ b/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
@@ -30,6 +30,10 @@ import {
   MessengerBotRequest,
   MessengerBotResponse,
   MessengerBotUser,
+  PersistentMenuItemType,
+  WebViewHeightRatio,
+  WebViewShareButton,
+  DisabledSurfaces,
 } from '.';
 
 export interface UpdateConfig<T> {
@@ -42,11 +46,30 @@ export interface GreetingElement {
   text: string;
 }
 
+export interface PersistentMenuItem {
+  type: PersistentMenuItemType;
+  title: string;
+  url?: string;
+  payload: string;
+  webview_height_ratio?: WebViewHeightRatio;
+  messenger_extensions?: Boolean;
+  fallback_url?: string;
+  webview_share_button?: WebViewShareButton;
+}
+
+export interface PersistentMenuElement {
+  locale: string;
+  composer_input_disabled?: boolean;
+  disabled_surfaces: Array<DisabledSurfaces>;
+  call_to_actions: Array<PersistentMenuItem>;
+}
+
 export interface Config extends ExtensibleConfig {
   shouldOverrideAppHandle?: boolean;
   shouldIgnoreSynchronousResponse?: boolean;
   greeting?: UpdateConfig<GreetingElement[]>;
   launch?: UpdateConfig<string>;
+  persistentMenu?: UpdateConfig<PersistentMenuElement[]>;
   pageAccessToken?: string;
   verifyToken?: string;
   locale?: string;
@@ -66,6 +89,9 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
       updateOnSetup: false,
     },
     launch: {
+      updateOnSetup: false,
+    },
+    persistentMenu: {
       updateOnSetup: false,
     },
     pageAccessToken: process.env.FB_PAGE_ACCESS_TOKEN || '',
@@ -161,6 +187,19 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
       }
 
       data.greeting = greetingElements;
+    }
+
+    if (this.config.persistentMenu && this.config.persistentMenu.updateOnSetup) {
+      const persistentMenuElements = this.config.persistentMenu.data;
+      if (!persistentMenuElements || persistentMenuElements.length < 1) {
+        throw new JovoError(
+          `Cannot set persistent menu elements to 'undefined' or an empty array.`,
+          ErrorCode.ERR_PLUGIN,
+          'FacebookMessenger',
+        );
+      }
+
+      data.persistent_menu = persistentMenuElements;
     }
 
     if (Object.keys(data).length === 0) {
--- a/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
+++ b/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
@@ -366,11 +366,12 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
         request.entry.length > 0;
       if (isMessengerRequest) {
         const promises: Array<Promise<any>> = [];
+        const responseObj: { message: MessengerBotResponse[] } = { message: [] };
         request.entry.forEach((entry: MessengerBotEntry) => {
           const hostCopy: Host = Object.create(host.constructor.prototype);
           // tslint:disable-next-line
           hostCopy.setResponse = async function (obj: any) {
-            return;
+            responseObj.message.push(obj.message);
           };
           for (const key in host) {
             if (host.hasOwnProperty(key)) {
@@ -389,7 +390,7 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
         });
 
         await Promise.all(promises);
-        return host.setResponse({});
+        return host.setResponse(responseObj);
       }
       return PROTOTYPE_BACKUP.call(this, host);
     };
--- a/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
+++ b/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
@@ -366,12 +366,12 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
         request.entry.length > 0;
       if (isMessengerRequest) {
         const promises: Array<Promise<any>> = [];
-        const responseObj: { message: MessengerBotResponse[] } = { message: [] };
+        const responses: MessengerBotResponse[] = [];
         request.entry.forEach((entry: MessengerBotEntry) => {
           const hostCopy: Host = Object.create(host.constructor.prototype);
           // tslint:disable-next-line
           hostCopy.setResponse = async function (obj: any) {
-            responseObj.message.push(obj.message);
+            responses.push(obj);
           };
           for (const key in host) {
             if (host.hasOwnProperty(key)) {
@@ -390,7 +390,7 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
         });
 
         await Promise.all(promises);
-        return host.setResponse(responseObj);
+        return host.setResponse(responses);
       }
       return PROTOTYPE_BACKUP.call(this, host);
     };
--- a/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
+++ b/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
@@ -369,10 +369,6 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
         const responses: MessengerBotResponse[] = [];
         request.entry.forEach((entry: MessengerBotEntry) => {
           const hostCopy: Host = Object.create(host.constructor.prototype);
-          // tslint:disable-next-line
-          hostCopy.setResponse = async function (obj: any) {
-            responses.push(obj);
-          };
           for (const key in host) {
             if (host.hasOwnProperty(key)) {
               const value = (host as any)[key];
@@ -386,6 +382,10 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
               }
             }
           }
+          // tslint:disable-next-line
+          hostCopy.setResponse = async function (obj: any) {
+            responses.push(obj);
+          };
           promises.push(PROTOTYPE_BACKUP.call(this, hostCopy));
         });
 
--- a/jovo-platforms/jovo-platform-facebookmessenger/src/core/MessengerBotResponse.ts
+++ b/jovo-platforms/jovo-platform-facebookmessenger/src/core/MessengerBotResponse.ts
@@ -20,9 +20,7 @@ export class MessengerBotResponse implements JovoResponse {
     return Object.assign(response, json);
   }
 
-  // Make message an array to handle cases where the bot is able to send
-  // multiple responses in one invocation.
-  message?: Message[];
+  messages: Message[] = [];
 
   getReprompt(): string | undefined {
     return undefined;
@@ -45,7 +43,15 @@ export class MessengerBotResponse implements JovoResponse {
   }
 
   getSpeechPlain(): string | undefined {
-    return this.message instanceof TextMessage ? this.message.message.text : undefined;
+    return this.messages
+      .map((m) => {
+        if (m instanceof TextMessage) {
+          return m.message.text;
+        }
+        return;
+      })
+      .filter((m) => !!m)
+      .join(' ');
   }
 
   hasSessionAttribute(name: string, value?: any): boolean {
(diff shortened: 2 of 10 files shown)


aswetlow:
Hey @brody-ly 
Thank you for the detailed PR. Those are valid feature proposals. We'll take a look after our v4 beta release.

jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
```diff
@@ -290,6 +330,13 @@ export class FacebookMessenger extends Platform<MessengerBotRequest, MessengerBo
     }
   }
 
+  async afterResponse(handleRequest: HandleRequest) {
+    if (!handleRequest.jovo || handleRequest.jovo.constructor.name !== this.getAppType()) {
+      return Promise.resolve();
+    }
+    await handleRequest.host.setResponse(handleRequest.jovo.$response);
```
m-ripper, line 337: I haven't tested this piece yet, but doesn't [this part](https://github.com/jovotech/jovo-framework/blob/7c5ca44703e483c45f40ced5dfc2f85cf5e43544/jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts#L357) cause the response to always be an empty object?

m-ripper (commented):
Thanks for the in-depth pull request and sorry for taking so long, we're working on V4. 

As far as I am aware, everything looks fine. I just left a comment regarding setting the responses. 

After we've resolved that, we should be able to merge your changes. 🚀

jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
brody-ly: Hey @m-ripper. Nice catch! That would definitely return an empty object for request body that takes the following format

```
{
  "object":"page",
  "entry":[
    {
      "id":"<PAGE_ID>",
      "time":1458692752478,
      "messaging":[
        {
          "sender":{
            "id":"<PSID>"
          },
          "recipient":{
            "id":"<PAGE_ID>"
          },

          ...
        }
      ]
    }
  ]
}
```

What would you suggest we do? Would an option be to remove that block of code completely and just return at the end of `overrideAppHandle()`?

m-ripper: The biggest issue here is, that theoretically, a single request from Facebook can contain multiple events. 

What we could do though is keep the `afterResponse()`-hook in order to set the response, but change the code in `overrideAppHandle()`.
 
 I would recommend doing the following:
 ```
     // ... other code above
      if (isMessengerRequest) {
        const promises: Array<Promise<any>> = [];
        const responses: MessengerBotResponse[] = [];
        request.entry.forEach((entry: MessengerBotEntry) => {
          const hostCopy: Host = Object.create(host.constructor.prototype);
          // tslint:disable-next-line
          hostCopy.setResponse = async function (obj: any) {
            return responses.push(obj);
          };
          for (const key in host) {
            if (host.hasOwnProperty(key)) {
              const value = (host as any)[key];
              if (key === '$request') {
                (hostCopy as any)[key] = entry;
              } else {
                (hostCopy as any)[key] =
                  typeof value === 'object' && value.constructor.name === 'Object'
                    ? { ...value }
                    : value;
              }
            }
          }
          promises.push(PROTOTYPE_BACKUP.call(this, hostCopy));
        });

        await Promise.all(promises);
        return host.setResponse(responses);
      }
      // ... other code beneath
 ```
 
I haven't tested this, but this should work.
 
The response will then always be an array though, but I think that makes sense because we can also receive multiple events.

If you have any more questions or feedback, just let me know.

brody-ly: Hey @m-ripper - Thanks for the suggestion! I see the issue you're pointing out about multiple events. I've tweaked the snippet of code you suggested and returned the same structure that I suggested in the PR. So the structure for when someone sends a request with multiple events would look something like this:

```
{
    "message": [
        [
            {
                "recipient": {
                    "id": "124"
                },
                "message": {
                    "text": "hello 1"
                },
                "message_type": "RESPONSE"
            }
        ],
        [
            {
                "recipient": {
                    "id": "124"
                },
                "message": {
                    "text": "hello 2"
                },
                "message_type": "RESPONSE"
            }
        ]
    ]
}
```

Let me know what you think about this

jovo-platforms/jovo-platform-facebookmessenger/test/FBMessengerResponse.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    app.handle(ExpressJS.dummyRequest(request));
+
+    // This hook should be called twice since there are two events are called
+    app.on('response', (handleRequest: HandleRequest) => {
+      const response = handleRequest.jovo!.$response;
+      const request: any = handleRequest.jovo?.$request;
+
+      const intentName = request.messaging[0].postback.payload;
+
+      if (intentName === 'INTENT_1') {
```
brody-ly, line 119: Hey @m-ripper - Any suggestions on how to improve the testing for a request with multiple events? I couldn't find a better way...

jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
m-ripper: First of all, thanks for the adjustments.

The problem I see with this structure is just that it's not really possible to tell which messages in `message` were created based on which event. 

Let's say two events were coming in, A and B, then it's not really easy to tell which of the messages belong to A and which belong to B, which might make testing harder. 

The implementation that I suggested solves this problem because response[0] will always contain the messages for event A and response[1] will always contain the messages for event B. 

For the example above, the response would then look like this:
```
[
  {
    "message": [
      {
        "recipient": {
          "id": "124"
        },
        "message": {
          "text": "hello 1"
        },
        "message_type": "RESPONSE"
      }
    ]
  },
  {
    "message": [
      {
        "recipient": {
          "id": "124"
        },
        "message": {
          "text": "hello 2"
        },
        "message_type": "RESPONSE"
      }
    ]
  }
]
```

But I am open to discuss this. Why would you prefer a single object? 

And another small thing I thought about is maybe renaming the `message` property in the response to `messages` because it's clearer that it is an array that way.

jovo-platforms/jovo-platform-facebookmessenger/test/FBMessengerResponse.test.ts
m-ripper: Shouldn't the response theoretically contain the messages of both events? In that case, matching the response against the expected response should be sufficient.

The expected response would contain both messages in this case.

m-ripper: That would be
```
{
  "message": [
    {
      "message": { "quick_replies": undefined, "text": "hello world 1" },
      "message_type": "RESPONSE",
      "recipient": { "id": "2321893584590078" }
    },
    {
      "message": { "quick_replies": undefined, "text": "hello world 2" },
      "message_type": "RESPONSE",
      "recipient": { "id": "2321893584590078" }
    }
  ]
}
            message: { quick_replies: undefined, text: 'hello world 1' },
              message_type: 'RESPONSE',
              recipient: { id: '2321893584590078' },    
    },
    {
                  message: { quick_replies: undefined, text: 'hello world 2' },
              message_type: 'RESPONSE',
              recipient: { id: '2321893584590078' },
    }
  ]
}
```
with the current structure I think.

m-ripper: Theoretically, you could dynamically check if the number in the text is the same as the number in the intent name via a regex. But I am not sure if this is cleaner from a testing perspective. Another option would be this:

```
    // Request will call INTENT_1 and INTENT_2 in one call
    const request: JovoRequest = await t.requestBuilder.launch(multipleEventRequest);
    const host = ExpressJS.dummyRequest(request);
    host.setResponse = function (response: any) {
      expect(response).toMatch({ message: [  ... ] }) // here the message-array should contain both messages
    }
    await app.handle(host);
```

Here, a test-Host is created and `setResponse` overwritten to match the response that is passed as the first argument. I haven't actually tested this, but I think this should work as well.

jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
brody-ly: Thanks for the explanation! I completely agree with what you are suggesting. I'll update the PR accordingly

jovo-platforms/jovo-platform-facebookmessenger/test/FBMessengerResponse.test.ts
brody-ly: I've tested the code above locally and found that `setResponse` gets called 3 times. The first 2 are the results of `INTENT_1` and `INTENT_2` and the last is strangely an empty array. The first two calls make sense, but the last call should theoretically contain the result of both intents in an array. However, the output of a running local server yields the correct result. Do you have any suggestions on how to get around the multiple calls during testing?

jovo-platforms/jovo-platform-facebookmessenger/src/FacebookMessenger.ts
m-ripper: Thank you very much!

jovo-platforms/jovo-platform-facebookmessenger/test/FBMessengerResponse.test.ts
m-ripper: Hey there, the reason why the response in `setResponse()` is empty, is due to overwriting the `setResponse()`-method set in `overrideAppHandle()`. The best solution to this is to set `setResponse()` in `overrideAppHandle()` after the merging took place:

```
          const hostCopy: Host = Object.create(host.constructor.prototype);
          for (const key in host) {
            if (host.hasOwnProperty(key)) {
              const value = (host as any)[key];
              if (key === '$request') {
                (hostCopy as any)[key] = entry;
              } else {
                (hostCopy as any)[key] =
                  typeof value === 'object' && value.constructor.name === 'Object'
                    ? { ...value }
                    : value;
              }
            }
          }
          // tslint:disable-next-line
          hostCopy.setResponse = async function (obj: any) {
            responses.push(obj);
          };
          promises.push(PROTOTYPE_BACKUP.call(this, hostCopy));
```
This will avoid similar errors in the future. 

The snippet above now looks like this and succeeded for me:
```
    const request: JovoRequest = await t.requestBuilder.launch(multipleEventRequest);
    const host = ExpressJS.dummyRequest(request);
    host.setResponse = async function (response: any) {
      expect(response).toEqual([
        {
          messages: [
            {
              message: { quick_replies: undefined, text: 'hello world 1' },
              message_type: 'RESPONSE',
              recipient: { id: '2321893584590078' },
            },
          ],
        },
        {
          messages: [
            {
              message: { quick_replies: undefined, text: 'hello world 2' },
              message_type: 'RESPONSE',
              recipient: { id: '2321893584590078' },
            },
          ],
        },
      ]); // here the message-array should contain both messages
      done();
    };
    await app.handle(host);
```

If you have questions, just let me know.

brody-ly: Thanks again @m-ripper! That fixed the issue 😄

jovo-platforms/jovo-platform-facebookmessenger/src/core/MessengerBotResponse.ts
```diff
@@ -42,7 +43,15 @@ export class MessengerBotResponse implements JovoResponse {
   }
 
   getSpeechPlain(): string | undefined {
-    return this.message instanceof TextMessage ? this.message.message.text : undefined;
+    return this.messages
```
brody-ly, line 46: Upon updating messages to be an array, I realised this method needed to change to reflect the new structure. I wasn't too sure what to do about this method but thought it might be good to return all FB texts as a string separated by space. What's your opinion on this?

m-ripper: Good catch! Yeah, I think that's good and should be sufficient. We will just have to update the docs that people know what to expect there.

brody-ly: That's a good point, I can add that in for sure. How would you prefer we add that configuration? I was thinking for simplicity's sake, we could just add a new attribute in the `MessengerBotResponse` class or we could update the `getSpeechPlain` signature to take a separator input?

m-ripper: I think being able to pass it as a parameter to `getSpeechPlain()` and making that parameter optional is probably the best option.

brody-ly: Sounds good, I've updated the PR.

m-ripper: Thanks you! I will check the changes once more and then we should be able to merge.

m-ripper (approved):
Thanks for the great work!

aswetlow:
Awesome, thanks! We will publish the update later today.

brody-ly:
Thanks guys! Appreciated all the help on this PR

PR #1002, opened 2021-08-24
♻️ Refactor input and routing

m-ripper:
## Proposed changes
- `nlu` and `asr` of `JovoInput` are now `undefined` by default.
- `nlu` of `JovoInput` is now only used by NLU-plugins and not platforms.
- `RouterPlugin` sets `intent` of `JovoInput` to the mapped intent if there is any and falls back to the intent otherwise. At this point, the `nlu.intent` of `JovoInput` will be used if there is no `intent` in the input.
- `RoutingExecutor` will look if the handler can handle the `type` of the input and otherwise looks if the handler can handle the `intent` of the input. At this point, it is also looked for `UNHANDLED`-handlers.
- Remove logic to overwrite `type` of the input in platforms.
- Smaller improvements

## Types of changes

## Checklist

### DIFF
--- a/docs/app-config.md
+++ b/docs/app-config.md
@@ -191,6 +191,33 @@ Especially with apps that work across different platforms, it might happen that 
 }
 ```
 
+For platforms like Alexa that already come with an intent in their requests, the mapped intent name is added to the root of the [`$input` object](./input.md):
+
+```typescript
+{
+  type: 'INTENT',
+  intent: 'HelpIntent',
+}
+```
+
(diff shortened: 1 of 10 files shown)


jankoenig (approved):
Thanks! I updated the docs. Good to go from my side ✅

rubenaeg (changes requested):
Looks good, could you resolve the linter warnings before merging?

rubenaeg (approved):

PR #1004, opened 2021-08-25
🚑️ switched from dot to index notation when filling entity map.

fbublitz:
## Proposed changes

Bugfix
Switch from dot to index notation when filling entity map.
The variable entityAlias is not evaluated when using dot notation, which results in always having the key "entityAlias" as the only key in the entity map instead of having the actual entity names as keys

## Types of changes

## Checklist


m-ripper (approved):

m-ripper:
Thank you very much, I must have missed this when reviewing.

jankoenig (approved):

PR #1009, opened 2021-08-27
 📝 Update docs

jankoenig:
This PR adds some overall improvements to the documentation, as well as new pages for user, models, Alexa CLI, Google Assistant CLI, and entities.

### DIFF
--- a/docs/entities.md
+++ b/docs/entities.md
@@ -1,0 +1,101 @@
+# Entities
+
+Learn more about how to access entities using Jovo.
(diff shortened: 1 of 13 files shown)


m-ripper (changes requested):
Besides the other comments, we should discuss whether if want to use relative paths instead of absolute paths in the .md-files in sub-folders. It should be cleaner and not be set to a hardcoded branch.

m-ripper (approved):

PR #1016, opened 2021-09-02
🏷️ 🐛 Fix rest-parameters for decorators

m-ripper:
## Proposed changes
The previous implementation caused the transpiled JS to not have a rest-parameter and therefore omit every argument after the first. Now a rest-parameter will always be assumed and the possible types that are passed are limited by overloaded signatures. In case an array is passed, some special treatment was necessary because the actual value will be an array inside the rest-parameter-array.

## Types of changes

## Checklist


aswetlow (approved):

jankoenig (approved):
Thanks! I just updated the docs. Good to go from my side ✅

PR #1022, opened 2021-09-08
👔 Update logic to skip redundant requests for GoogleBusiness

m-ripper:
## Proposed changes
Just like before, typing-indicator- and receipt-requests will be ignored, but now no further middlewares will be executed. This way the debugger for example will not log these requests.

Also improved `run` of `MiddlewareCollection` to check if the middleware exists after the `before.`-middleware has been executed. This way, removing middlewares in the `before.`-middleware allows skipping any further middlewares even the same middleware without the prefix.

## Types of changes

## Checklist


jankoenig (approved):
Thanks! I added documentation for `stopMiddlewareExecution`. Please take a look if I missed anything. Good to go from my side ✅

PR #1027, opened 2021-09-13
✨ Add accessToken property to JovoUser

aswetlow:
Adds `accessToken` property to `JovoUser` class + Implementations for `AlexaUser` and `GoogleAssistantUser`

`this.$user.accessToken` can be undefined or of type `string`. 

### Usage

```typescript
LAUNCH() {
    const accessToken = this.$user.accessToken;
    return this.$redirect(LoveHatePizzaComponent);
  }
```

In `v3` the `getAccessToken()` method was part of the abstract `JovoRequest` class. 

## Types of changes

## Checklist

### DIFF
--- a/platforms/platform-alexa/src/AlexaUser.ts
+++ b/platforms/platform-alexa/src/AlexaUser.ts
@@ -24,6 +24,10 @@ export class AlexaUser extends JovoUser<Alexa> {
     return this.jovo.$request.session?.user?.userId || 'AlexaUser';
   }
 
+  get accessToken(): string | undefined {
+    return this.jovo.$request.session?.user.accessToken;
+  }
+
   async getEmail(): Promise<string | undefined> {
     const request: AlexaRequest = this.jovo.$request;
     const email: string = await sendCustomerProfileApiRequest(
--- a/docs/user.md
+++ b/docs/user.md
@@ -58,6 +58,7 @@ const score = this.$user.data.score;
 Additionally to the persisted data, you can also access the following information about the user:
 
 * `this.$user.id`: The user's ID is also the key to their database entry.
+* `this.$user.accessToken`: For platforms that offer account linking, the `accessToken` is a string for signed in users. The value is `undefined` if the user hasn't linked their account, or the platform does not support account linking.
 * `this.$user.createdAt`: When was this user's database entry created?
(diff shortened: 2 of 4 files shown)


rubenaeg (approved):

jankoenig (approved):
Thanks! Just added documentation for this. Let me know if I missed anything.

platforms/platform-alexa/src/AlexaUser.ts
```diff
@@ -24,6 +24,10 @@ export class AlexaUser extends JovoUser<Alexa> {
     return this.jovo.$request.session?.user?.userId || 'AlexaUser';
   }
 
+  get accessToken(): string | undefined {
+    return this.jovo.$request.session?.user.accessToken;
```
rubenaeg, line 28: If `this.jovo.$request.session` is undefined, then `user` will be undefined too, probably resulting in an error such as this one:
```typescript
TypeError: Cannot read property 'accessToken'  of undefined
```
This can be solved by adding a question mark to `user` as well.

rubenaeg (changes requested):

platforms/platform-alexa/src/AlexaUser.ts
aswetlow: In my understanding. If `this.jovo.$request.session` is undefined, it instantly returns `undefined`. `user` is irrelevant for an undefined `$session`.

```typescript
interface FooBar {
  foo?: {
    bar: string;
  };
}

const obj: FooBar = {};

console.log(obj.foo?.bar);

```

Result:
```
undefined
```

rubenaeg: Sorry, I messed my initial comment up. What I meant is that if `this.jovo.$request.session` is defined, but `user` is not, then it will throw said error. We check for this exact thing in the getter for `AlexaUser.prototype.id` above, that's why I mentioned it. If we can safely assume for `user` to be always defined, this conversation can be resolved.

aswetlow: https://github.com/alexa/alexa-apis-for-nodejs/blob/009292ae927b7ac587d9eeb108a7134930d9a48c/ask-sdk-model/index.ts#L668

`user` is not optional.

PR #1029, opened 2021-09-16
✨ Implement handling of potential duplicate messages

m-ripper:
## Proposed changes
Like described [here](https://developers.google.com/business-communications/business-messages/guides/learn/best-practices#check_for_duplicate_incoming_messages), it is important to check whether the `messageId` was already handled and ignore the request in that case.

## Types of changes

## Checklist

### DIFF
--- a/platforms/platform-googlebusiness/src/GoogleBusinessPlatform.ts
+++ b/platforms/platform-googlebusiness/src/GoogleBusinessPlatform.ts
@@ -126,6 +126,9 @@ export class GoogleBusinessPlatform extends Platform<
       (plugin) => plugin instanceof DbPlugin,
     ) as DbPlugin[];
 
+    // Immediately save the data into the database, so that other simultaneous or delayed requests can already check if the message is being handled or was handled already.
+    // If some time-consuming API calls were made during the handling, the saving of the processed message would only take place after those calls which could cause a delay.
+    // This delay could then cause the persisting to happen after other requests have already checked if the message was handled.
     return Promise.all(dbPlugins.map((dbPlugin) => dbPlugin.saveData(jovo)));
   }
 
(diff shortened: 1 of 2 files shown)


platforms/platform-googlebusiness/src/GoogleBusinessPlatform.ts
```diff
@@ -97,6 +103,32 @@ export class GoogleBusinessPlatform extends Platform<
...
+    const processedMessages = jovo.$session.data._GOOGLE_BUSINESS_PROCESSED_MESSAGES_ || [];
+    // message was handled already
+    if (processedMessages.includes(messageId)) {
+      jovo.$handleRequest.stopMiddlewareExecution();
+      return jovo.$handleRequest.server.setResponse({});
+    }
+    processedMessages.push(messageId);
+    jovo.$session.data._GOOGLE_BUSINESS_PROCESSED_MESSAGES_ = processedMessages;
+
+    const dbPlugins = Object.values(jovo.$handleRequest.plugins).filter(
```
jankoenig, line 125: Quick question for me to understand this: I was under the assumption that the session data is stored to the DB anyways (happening later in the lifecycle). Could you tell me why we need to access the DB plugin here? Thanks!

m-ripper: Like described in this [commit](https://github.com/jovotech/jovo-framework/commit/e6ad0633d14574ecb5b5a0468c87d304f03f0c68):

> An immediate save is required because a Lambda may start multiple instances which then need to be aware of the request from the other instance.

Basically, we want other instances that start a little bit delayed to be able to check if the message was handled or is being handled. This is an edge-case probably but should be taken in consideration.

jankoenig: Cool, thanks. Could you add this as a comment?

jankoenig (approved):
Thanks! Approved ✅

aswetlow (approved):

PR #1037, opened 2021-09-21
♻️ Improve response for Facebook Messenger

m-ripper:
## Proposed changes
The response is now an array where each element is a response related to the processed event.

Example:
1. Request with a single event and text 'hello' received -> processed by the app -> The response is an array with a single item representing the response for the text 'hello'

2. Request with two events received (I) text 'hello' (II) text 'bye' -> I and II processed -> The response is an array with the first element representing the response for the text 'hello' and the second element representing the response for the text 'bye'.

Related to the same problem in #983.

## Types of changes

## Checklist


jankoenig:
Thanks! One quick question: How would this look like for serverless environments like AWS Lambda? e.g. the first event 'hello' would be handled by a different instance than the second event 'bye'. My assumption is that this would still work without a problem, the bot would still return 2 responses. Just wanted to keep this in mind as well

jankoenig (approved):
Just added some docs. Let me know if I missed anything. Thanks!

aswetlow (approved):

m-ripper:
> Thanks! One quick question: How would this look like for serverless environments like AWS Lambda? e.g. the first event 'hello' would be handled by a different instance than the second event 'bye'. My assumption is that this would still work without a problem, the bot would still return 2 responses. Just wanted to keep this in mind as well

In Lambda it should also just open a single instance, which then handles both events because just a single invocation happens.

jankoenig:
Ah got it, thank you. I misread and thought it was about multiple requests, not multiple events in a single request. I couldn't find anything in the Facebook docs regarding this behavior, do you have a link handy? Thanks

jankoenig:
Found it! Ready to merge

PR #1102, opened 2021-11-11
👔 Do not close application on debugger connection error

m-ripper:
## Proposed changes
- The debugger does not cause the process to exit anymore in case a connection error occurred.
- The debugger connection error is only shown once, but can be shown again after the socket could connect and disconnected again. Otherwise, the connection error would be spammed. Multiple error messages can still occur if multiple requests are processed simultaneously and the socket endpoint closes.

## Types of changes

## Checklist


jankoenig (approved):
Ah, that's a good one. Thanks!

aswetlow (approved):

PR #1105, opened 2021-11-12
:label: Fix typings

rubenaeg:
## Proposed changes
This PR fixes Linter issues and introduces `PartialWhere` to eliminate the use of duplicate types in `@jovotech/platform-alexa`.

## Types of changes

## Checklist

### DIFF
--- a/framework/src/App.ts
+++ b/framework/src/App.ts
@@ -198,13 +198,15 @@ export class App extends Extensible<AppConfig, AppMiddlewares> {
     }
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async handleError(e: Error, jovo?: Jovo): Promise<void> {
+  async handleError(error: unknown, jovo?: Jovo): Promise<void> {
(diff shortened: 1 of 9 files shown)


aswetlow (approved):

jankoenig (approved):

framework/src/App.ts
```diff
@@ -199,7 +199,7 @@ export class App extends Extensible<AppConfig, AppMiddlewares> {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async handleError(e: any, jovo?: Jovo): Promise<void> {
+  async handleError(e: Error, jovo?: Jovo): Promise<void> {
```
m-ripper, line 202: I am not sure if we should change the type because it is theoretically possible that the value that is passed there is not an `Error`.

rubenaeg: Would `unknown` be the better choice here?

m-ripper: I agree, `unknown` would be the better choice.

m-ripper (approved):

PR #1140, opened 2021-11-26
:sparkles: Introduce Alexa Conversations

rubenaeg:
## Proposed changes
This PR introduces Alexa Conversations, along with improvements to the deploy process of an Alexa skill.

## Types of changes

## Checklist


aswetlow (approved):

platforms/platform-alexa/src/output/templates/ApiResponseOutput.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface ApiResponseOutputOptions extends OutputOptions {
+  apiResponse: Record<string, string | number | boolean>;
+}
+
+@Output()
+export class ApiResponseOutput extends BaseOutput<ApiResponseOutputOptions> {
+  build(): OutputTemplate | OutputTemplate[] {
+    return {
+      listen: this.options.listen,
```
jankoenig, line 11: `listen` is a [reserved property](https://www.jovo.tech/docs/output-classes#reserved-properties), so it doesn't necessarily need to be added to this output. By default, it's `true`

platforms/platform-alexa/src/output/templates/ApiResponseOutput.ts
```diff
@@ -0,0 +1,25 @@
...
+      platforms: {
+        alexa: {
+          nativeResponse: {
+            version: '1.0',
```
jankoenig, line 15: `version` and `sessionAttributes` isn't needed here, will be added automatically

platforms/platform-alexa/src/output/templates/ApiResponseOutput.ts
jankoenig: never mind, I think this is OK and clearer if we keep it in there

platforms/platform-alexa/src/output/templates/ApiResponseOutput.ts
jankoenig: let's keep it in there for now

jankoenig (approved):

PR #1146, opened 2021-12-02
✨♻️ Implement LexSlu and generic AsrPlugin and SluPlugin classes 

m-ripper:
## Proposed changes
- Implement `AudioUtilities` and `ParsedAudioInput` for easier handling of audio
- Implement `getText` method for `JovoInput` which checks for the `asr.text`. Every usage of `$input.text` has been replace with `getText()`
- Implement `InterpretationPlugin` which contains logic for asr- and nlu-processing. Now `NluPlugin` extends `InterpretationPlugin` and `AsrPlugin` as well as `SluPlugin` that extend `InterpretationPlugin` have been implemented as well for their respective tasks. The advantage of this structure is that changes can be made across different tasks -> less duplicate code and easier maintainability
- Remove unused dependencies

TODO: The base64 audio string has to be excluded/replaced everywhere when logging, otherwise it will clutter the logs. This is probably going to happen in a separate PR though.

## Types of changes

## Checklist


aswetlow (approved):

integrations/slu-lex/src/LexSlu.ts
```diff
@@ -0,0 +1,170 @@
...
+    const nluData: NluData = {
+      intent: interpretation.intent.name,
+    };
+    if (interpretation.intent.slots) {
+      nluData.entities = Object.entries(interpretation.intent.slots).reduce(
+        (entities: EntityMap, [name, slot]) => {
+          if (!slot.value) {
+            return entities;
+          }
+          entities[name] = {
```
jankoenig, line 155: Is it possible that `id` is missing here? Should be the same as `resolved` if Lex doesn't offer an `id`

m-ripper: Thanks for pointing that out. The `id` is now properly set as well.
Lex sadly doesn't have a specific id, therefore it will be the same value as `resolved`.

jankoenig (approved):

PR #1148, opened 2021-12-08
🐛 Fix bug related to component config

m-ripper:
## Proposed changes
- Fix bug that caused config which was set in the decorator to not be applied.
- The options of the component can now be retrieved by using `this.options` from a component

## Types of changes

## Checklist

### DIFF
--- a/framework/src/BaseComponent.ts
+++ b/framework/src/BaseComponent.ts
@@ -1,17 +1,16 @@
-import { UnknownObject } from '@jovotech/common';
+import { DeepPartial, UnknownObject } from '@jovotech/common';
 import { ComponentData, JovoComponentInfo } from './index';
 import { Jovo } from './Jovo';
 import { JovoProxy } from './JovoProxy';
 import { ComponentOptions } from './metadata/ComponentMetadata';
 
-export type ComponentConstructor<COMPONENT extends BaseComponent = BaseComponent> = new (
-  jovo: Jovo,
-  options?: ComponentOptions<COMPONENT>,
-  ...args: unknown[]
-) => COMPONENT;
-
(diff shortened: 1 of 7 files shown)


aswetlow (approved):

framework/src/BaseComponent.ts
```diff
@@ -1,18 +1,17 @@
...
 import { Jovo } from './Jovo';
 import { JovoProxy } from './JovoProxy';
 import { ComponentOptions } from './metadata/ComponentMetadata';
 
+export type ComponentConfig<COMPONENT extends BaseComponent = BaseComponent> =
+  COMPONENT['$component']['config'];
+
 export type ComponentConstructor<COMPONENT extends BaseComponent = BaseComponent> = new (
   jovo: Jovo,
-  options?: ComponentOptions<COMPONENT>,
```
jankoenig, line 9: I'm not sure if I'm mistaken, but shouldn't we allow to pass all `options` to the constructor, not just the `config`?

m-ripper: `options` refers to the data that can be passed to the `@Component`-decorator.
We only need the configuration when initializing the component itself, the other properties are not used by the component instance itself. But theoretically we could pass the options as well.

jankoenig: Thanks. I'm still not sure. Isn't `name` an important part of the `options` as well?

```
components: [MenuComponent, new ComponentDeclaration(MenuComponent2, { name: 'MenuComponent2' })],
```

Or is this still possible?

m-ripper: Good catch, in some edge cases it could be actually helpful to have access to the other options. I will change it accordingly.

jankoenig (approved):

PR #1152, opened 2021-12-10
:bug: Fix Task config property

rubenaeg:
## Proposed changes
This PR fixes a small bug where `indent` was used in Task initialization instead of the correct `indentation`.

## Types of changes

## Checklist


jankoenig (approved):

integrations/plugin-debugger/src/JovoDebugger.ts
```diff
@@ -464,15 +464,6 @@ export class JovoDebugger extends Plugin<JovoDebuggerConfig> {
     await this.emitDebuggerConfig();
     await this.emitLanguageModelIfEnabled();
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    function propagateStreamAsLog(stream: Writable, socket: typeof Socket) {
```
m-ripper, line 469: Please add this function again. It is not used now but might be used in the future.

m-ripper (changes requested):
Beside that one change in the Debugger, it looks good. 
Did you test the unknown typing? Due to unknown being stricter further error could occurre in integrations or apps.

integrations/plugin-debugger/src/JovoDebugger.ts
rubenaeg: Added it in `utilities.ts` :+1:

m-ripper: Can you remove the `@ts-ignore` please? Casting would be a much cleaner solution if necessary.

PR #1175, opened 2022-01-06
[PROPOSAL] add `RichAudio` output type

jtfell:
Adds a new type of output to handle response types for combined audio, silence and TTS content. The platforms handle this quite differently:

1. [Alexa APLA](https://developer.amazon.com/en-US/docs/alexa/alexa-presentation-language/apla-interface.html)

This is a straightforward mapping (I used their `item` structure to keep things simple here).

2. [Google Assistant SSML](https://cloud.google.com/text-to-speech/docs/ssml)

Rich audio is represented with non-standard SSML tags in the `speech` response. 

## Proposed changes

- add a generic `RichAudio` model to the output package that is modelled on the APLA data structure.
- add support for rendering it equivalently on both the `alexa` and `googleAssistant` platforms.
- multiple `richAudio` outputs would be merged into a single response by wrapping them in a `Sequencer` item so they are played sequentially

Example usage:

```typescript
this.$send({
  richAudio: {
    type: 'Sequencer',
    items: [
      {
        type: "Speech",
        content: "Text content"
      },
      {
        type: "Audio",
        source: "https://assets.com/intro_sound.mp3"
      }
    ]
  }
})
```

On Google Assistant, this example would be converted to:

```xml
<seq>
    <media><p>Text content</p></media>
    <media><audio src="https://assets.com/intro_sound.mp3" /></media>
</seq>
```

On Alexa, this example would be:

```json
{
    "type": "Alexa.Presentation.APLA.RenderDocument",
    "token": "token",
    "document": {
        "type": "APLA",
        "version": "0.8",
        "mainTemplate": {
            "parameters": ["payload"],
            "item": {
                "type": "Sequencer",
                "items": [
                    {
                      "type": "Speech",
                      "content": "Text content"
                    },
                    {
                      "type": "Audio",
                      "source": "https://assets.com/intro_sound.mp3"
                    }
                ]
            }
        }
    },
    "datasources": {}
}
```

## Types of changes

## Checklist


jtfell:
Any ideas why that function is undefined when running the tests?

https://github.com/jovotech/jovo-framework/runs/4732801092?check_suite_focus=true

jtfell:
Can I get you to take a look at this @jankoenig? Really keen for some feedback on the design

jankoenig:
Hi @jtfell, I'm sorry this is taking so long. The reason is that I'm not sure what's the best design for something like this. We don't want to add too many generic properties that only work for a few platforms. This is why I'm currently thinking how this could look like for other platforms. Thanks for your patience!

jtfell:
I totally understand, I agree it's worth getting the design right and we're not in a big rush to get anything merged. I'm just keen to get the conversation started on how to best achieve this with Jovo

From our experience with voice apps, we have found the mixing of audio files (for sound effects, branding etc) with text-to-speech responses as being really core to a good experience on Alexa and Google Assistant. I'm not at all sure that this is the best way to support this kind of thing in Jovo, but our experience with Jovo v3 was that it was pretty messy trying to achieve it without functionality built into the framework.

jankoenig:
> I totally understand, I agree it's worth getting the design right and we're not in a big rush to get anything merged. I'm just keen to get the conversation started on how to best achieve this with Jovo
> 
> From our experience with voice apps, we have found the mixing of audio files (for sound effects, branding etc) with text-to-speech responses as being really core to a good experience on Alexa and Google Assistant. I'm not at all sure that this is the best way to support this kind of thing in Jovo, but our experience with Jovo v3 was that it was pretty messy trying to achieve it without functionality built into the framework.

Yes, I agree. I'm thinking about adding a bit more generic `audio` property instead of `richAudio`. Let me get back to you after we release `4.2` this week

jtfell:
I've added some more docs, and played around with a merging strategy that I think plays nicer alongside the `message` output property.

jankoenig:
Thank you @jtfell! It's a great idea, but we're closing this for now because we're struggling a bit to find a consistent way to integrate this. We'll gladly pick it up again if the topic gets prioritized in the future.

PR #1190, opened 2022-01-19
💥♻️ Refactor `JovoLogger`

m-ripper:
## Proposed changes
- Refactor `JovoLogger` to use [loglevel](https://github.com/pimterry/loglevel) instead of [tslog](https://tslog.js.org/#/) due to issues with browser-compatibility when using tslog.
- Allow a string to be passed as constructor parameter `JovoError` to pass a message only
- The properties and order of properties of `JovoError` that should be listed when printing a `JovoError` with the `JovoLogger`  can be modified by passing `properties` to the config of `JovoLogger`. 
- Remove tslogMockPlugin from web-client build-script

> Before merging we will have to discuss how the current `properties`-property in `JovoLoggerConfig` is supposed to be named.

## Types of changes

## Checklist

### DIFF
--- a/common/src/JovoLogger.ts
+++ b/common/src/JovoLogger.ts
@@ -1,6 +1,6 @@
 import chalk, { Chalk } from 'chalk';
-import _merge from 'lodash.merge';
-import { levels, getLogger, Logger, LogLevelDesc } from 'loglevel';
+import _mergeWith from 'lodash.mergewith';
+import { getLogger, levels, Logger, LogLevelDesc } from 'loglevel';
 import { JovoError } from './JovoError';
 
 export interface JovoLoggerConfig {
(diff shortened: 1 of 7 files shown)


common/src/JovoLogger.ts
```diff
@@ -1,120 +1,175 @@
...
-import { AnyObject } from './index';
-export class JovoLogger extends TsLogger {
-  constructor(settings?: ISettingsParam) {
-    super();
-    this.setSettings(_merge(this.settings, this.getDefaultConfig(), settings));
+
+export interface JovoLoggerConfig {
+  name: string | symbol;
+  level: LogLevelDesc;
+  disableStyling: boolean;
```
jankoenig: I'm not 100% sure on the naming of that property.  `logging` includes a property called `styling`. `logger` uses `disableStyling`. I think we should be consistent here

common/src/JovoLogger.ts
```diff
@@ -1,120 +1,175 @@
...
+  // TODO determine name
+  properties: Array<keyof JovoError>;
```
jankoenig: Have we finalized that property name yet?

jankoenig (changes requested):
I've just updated the docs a bit.

Could you change `properties` to `errorProperties`? After that, I think this is ready to merge 👍🏻  Thanks!

aswetlow (approved):

jankoenig (approved):

Issue #1176, opened 2022-01-08
Message speech is trimmed on Google Assistant


acerbisgianluca:
## I'm submitting a...

- [x] Bug report
- [ ] Feature request
- [ ] Other... Please describe:
- [ ] Documentation issue or request

## Expected Behavior

Speech hasn't a characters limit on Google AFAIK. Just the text has a limit of 640 characters. So I expect the text to be trimmed, but not the speech property.

Reference: [Google docs](https://developers.google.com/assistant/conversational/prompts-simple#SimpleResponseProperties)

## Current Behavior

The speech property is sanitized here:
https://github.com/jovotech/jovo-framework/blob/05f0a4b7308a358a9eaa8de3535773b6419e1d8e/platforms/platform-googleassistant/src/output/GoogleAssistantOutputTemplateConverterStrategy.ts#L41
and trimmed here:
https://github.com/jovotech/jovo-framework/blob/05f0a4b7308a358a9eaa8de3535773b6419e1d8e/output/src/OutputTemplateConverterStrategy.ts#L157

### Error log
```
$output.message was trimmed due to exceeding the limit of 640 characters.
```

## Your Environment

* Jovo Framework version used: 4.1.0
* Operating System: Windows 10

acerbisgianluca:
The default max length is set to `TEXT_MAX_LENGTH`:
https://github.com/jovotech/jovo-framework/blob/05f0a4b7308a358a9eaa8de3535773b6419e1d8e/platforms/platform-googleassistant/src/output/GoogleAssistantOutputTemplateConverterStrategy.ts#L59-L64

Here, calling `sanitizeMessage` with a much higher value for the maxLength parameter, should fix this issue:
https://github.com/jovotech/jovo-framework/blob/05f0a4b7308a358a9eaa8de3535773b6419e1d8e/platforms/platform-googleassistant/src/output/GoogleAssistantOutputTemplateConverterStrategy.ts#L41

aswetlow:
You're right. 

`speech` and `text` should be sanitized separately.

cc  @m-ripper

m-ripper:
This issue should be fixed with the update that is coming today.

Closing for now. If it still occurs, feel free to re-open.

PR #1192, opened 2022-01-20
👔 Implement separate sanitization of `text` and `speech`

m-ripper:
## Proposed changes
Currently, only a single max length can be set for `message` and `reprompt` which is then applied to both `text` and `speech` if they exist.

Now, an object can also be passed to the `maxLength` parameter of `OutputTemplateConverterStrategy.sanitizeMessage` for more control over separate max lengths for `text` and `speech`:
```
{
  speech?: number;
  text?: number;
}
```
If only a string is found as message, it will be handled as `speech` if SSML tags are found, otherwise it is treated as `text`.

`GoogleAssistantOutputTemplateConverterStrategy` overrides `normalizeOutput` to make Message-objects out of every string in `message` and `reprompt` before calling the parent's `normalizeOutput` method. \
This is done because later, a string will be converted to an object with `speech` and `text` anyways and this way the object will also get properly sanitized.

Related to #1176

## Types of changes

## Checklist

### DIFF
--- a/output/src/OutputTemplateConverterStrategy.ts
+++ b/output/src/OutputTemplateConverterStrategy.ts
@@ -5,6 +5,7 @@ import {
   Carousel,
   DynamicEntities,
   DynamicEntityMap,
+  isSSML,
   MessageValue,
   NormalizedOutputTemplate,
   OutputTemplate,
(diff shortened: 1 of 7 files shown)


aswetlow (approved):

jankoenig (approved):
Thanks! Just added some docs ✅

PR #1209, opened 2022-01-31
Reduce minimum items for carousel to 1

jtfell:
## Proposed changes

We are using carousels for a core part of the visuals in a Alexa and Google Assistant app, and have found the 2 item minimum on carousels to be obstructing the UX. On Alexa, item lists can contain a single item and we'd like to use them in that way for consistency as we supply options via voice. 

My understanding (correct me if I'm wrong) is that the restriction is there because Google Assistant has this requirement:
[Collections must contain a minimum of 2 and a maximum of 10 tiles. On display-capable devices, users can swipe left or right to scroll through cards in a collection before selecting an item.](https://developers.google.com/assistant/conversational/prompts-selection#collection)

I propose reducing the minimum items to 1 and displaying a [card](https://developers.google.com/assistant/conversational/prompts-rich#basic_card) on Google Assistant in this situation.

## Types of changes

## Checklist


jankoenig:
Thank you @jtfell! This makes a lot of sense.

The only concern I had was that we need to be consistent across all platforms here. I checked all of them:

- Facebook Messenger: Not a problem, here the minimum array size for generic templates is `1` anyway
- Google Business Messages: Here, we would also need a similar solution to Google Assistant, since the minimum amount of items in a `carousel` is `2`: https://developers.google.com/business-communications/rcs-business-messaging/guides/build/messages/send#carousels

jtfell:
Good catch @jankoenig - my head is almost entirely caught up with Alexa and Google Assistant at the moment. I've added a similar solution for Google Business (based on the docs) but don't have a place to test it as I'm not set up on that platform.

jankoenig (approved):
Thank you @jtfell, great PR 🎉 

I took this as an opportunity to improve the output docs everywhere and also add a sample project for Google Business Messages.

jtfell:
The updated docs look great 👍

aswetlow (approved):

jankoenig:
Merged. Thanks a lot @jtfell, great PR 🎉

PR #1212, opened 2022-02-01
Fix bug in handling of APL user events

jtfell:
## Proposed changes

Buttons pressed in an APL carousel were not working with the existing implementation. It also feels cleaner having the handling of this special request type in the `AlexaRequest.ts` module.

## Types of changes

## Checklist

### DIFF
--- a/platforms/platform-alexa/docs/output.md
+++ b/platforms/platform-alexa/docs/output.md
@@ -166,6 +166,17 @@ This will add the `intent` and `entities` properties to the APL document as `arg
 
 These are then mapped correctly and added to the [`$input` object](https://www.jovo.tech/docs/input) when the user taps a button.
 
+In your [handler](https://www.jovo.tech/docs/handlers), you can then access the entities like this:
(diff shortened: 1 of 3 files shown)


jankoenig (approved):
Thank you! 🎉 

I took this as an opportunity to add an APL section to the docs.

jtfell:
The docs look good @jankoenig! My only suggestion would be a more concrete example of using the selection in a handler. Eg.

```json
"arguments": [
  {
    "type": "Selection",
	  "intent": "ElementIntent",
	  "entities": {
      "element": {
        "value": "A"
      }
    }
  }
]
```

can be handled like this:

```ts
@Intents(['ElementIntent'])
HandleElementIntent() {
  const selection = this.$input.entities.element.value;
}
```

jankoenig:
Makes sense! Thank you

aswetlow (approved):

jankoenig (approved):

Issue #1214, opened 2022-02-02
MongoDB integration


jrglg:
## I'm submitting a...

- [ ] Bug report
- [X] Feature request
- [ ] Documentation issue or request
- [ ] Other... Please describe:

## Feature requests

MongoDb integration would be nice. Many advantages over Dynamodb, such as **being able to search JSON fields easily** and more flexibility in model changes.

In [MongoDB Atlas docs](https://docs.atlas.mongodb.com/best-practices-connecting-from-aws-lambda/) there is one point that is important: Create the MongoClient object once.

Thank you.

jankoenig:
Thank you. This is not on our immediate roadmap and would be a great community contribution!

If anyone wants to get started with this, let us know.

Two resources that should be helpful for this:
- The `v4` integration for DynamoDb, which can serve as a good example for getting started: https://github.com/jovotech/jovo-framework/tree/v4/latest/integrations/db-dynamodb
- The `v3` integration for MongoDB: https://github.com/jovotech/jovo-framework/tree/v3/latest/jovo-integrations/jovo-db-mongodb

jrglg:
I'm not clear on how to handle the instance of the connection.

According to [this](https://docs.atlas.mongodb.com/best-practices-connecting-from-aws-lambda/#connection-examples), possible deployment in Lambda should be considered. As it was in version 3, several connections were opened and it was not ideal.

The problem is that I don't know much about the language, and I'm not sure about the following:

1. whether we should provide the connection explicitly through an access like "this.$db". To me this makes sense because Jovo will open a connection to update the date of users last connection, and possible changes to "this.$user.data"... If, in addition, another use case requires access to the database, the connection should be reused.

2. How is this handled across different AWS Lambda invocations? As I understand it, a lot of benefit could be gained by initiating the connection outside the handler.

3. I'm not clear how to maintain the connection in Node.js. I don't know if it could work with a Singleton.
 Would option 1 be better? or you could do it directly from the MongoDb module. Would the latter work?

I'm already coding this but advise would be great.

jrglg:
Hi @jankoenig @aswetlow 

I found a weird issue when trying to use the helper methods under `this.$mongoDb`:

```typescript
  async LAUNCH() {
    const db= await this.$mongoDb.getDb(); // throws error
    const users = await this.$mongoDb.getCollection(); // throws error
    const client= await this.$mongoDb.getClient(); // throws error
```

The error is:

```
[...]

PR #1217, opened 2022-02-03
Add MongoDB integration

jrglg:
## Proposed changes
Add MongoDB integration.

Seems to work fine. I save the connection promise as said in #1214

## Types of changes

## Checklist


integrations/db-mongodb/src/MongoDb.ts
```diff
@@ -0,0 +1,119 @@
...
+      connectionString: '<YOUR-MONGODB-URI>',
+    };
+  }
+
+  mount(parent: HandleRequest): Promise<void> | void {
+    super.mount(parent);
+  }
+
+  async initialize(): Promise<void> {
+    this.config.databaseName = this.config.databaseName;
```
aswetlow: I think lines 58 and 58 can be removed.

jrglg: Oops. I would like the connectionPromise to be accesible from any component.
How could it be done?

jrglg:
As you can see I used the Singleton pattern in order to minimize connections. Is it weird for the current project estructure or does it suit well?
I'm a little concern as some projects with lots of users (like the one I'm developing for my University) could experience some issue with this if it's not well optimized. Also, I think it's the way to go.

In any case, I'm open to change everything you require

jankoenig:
Thank you @jrglg! This week was a bit hectic, but we're going to review all open PRs next week

integrations/db-mongodb/README.md
```diff
@@ -0,0 +1,101 @@
...
+Add it as plugin to any stage you like, e.g. `app.prod.ts`:
+
+```typescript
+import { MongoDb } from '@jovotech/db-mongodb';
+
+// ...
+
+app.configure({
+  plugins: [
+    MongoDb.instance({
```
jankoenig: I would prefer `new MongoDb({ /* config */ })` here because it's our convention for Jovo plugins.

Could you tell us a bit more why you chose that approach?

jankoenig (changes requested):

integrations/db-mongodb/README.md
jrglg: Sure. I did it that way because according to MongoDB best practices:

```
Define the client to the MongoDB server outside the [AWS Lambda handler function](http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html).

Don't define a new MongoClient object each time you invoke your function. Doing so causes the driver to create a new database connection with each function call. This can be expensive and can result in your application exceeding database connection limits. As an alternative, do the following:

Create the MongoClient object once.
Store the object so your function can reuse the MongoClient across function invocations.
```
It's a way to keep the same connection pool across different invocations. AWS Lambda may reuse the same connection too if the container is not killed in successive invocations.

Any operation on the DB would use the same connection. For example:
1. When Jovo saves data in the lifecycle through assignments in jovo.$user.data.
2. When the user needs to query another MongoDB DB, another collection, or any database data they would not have to create a new instance (and another connection). It would be reused.


It's the only way I thought. Other ways I came up with would require the user to add more configuration, and generally scramble the code to achieve the same thing.

jrglg: By the way, The package is published for testing: https://www.npmjs.com/package/@jorgelg/mongodb

jankoenig: Thank you. 

Or maybe we could offer a method like `this.$db.getInstance()` or `this.$mongoDb.getInstance()`?

jrglg: Yes, `this.$db.getInstance()` would be great.

I understand that the object would be created with 'new' like any other plugin, and from the accessor method the instance would be retrieved, returning something like `this.plugins.mongoDb`.

Actually though, I'm thinking that stages having two databases (rare, but could happen) would pose a problem calling `this.$db` because it wouldn't know which one to return.

It may be better to use `this.$mongoDb`

jrglg: I'm taking a look now. Could you please tell me how to get the MongoDB instance from App inside Jovo.ts?

jrglg: Another aproach would be:

1. use `this.mongoDb()`, because `this.$db` is more complicated for me, more checks would have to be done and 
and I can't think of an elegant solution.
2. a new interface in `DbPlugin` that looks something like this:
```
export interface DbInstance {
  getClient(): unknown;
  getJovoDb(): unknown;
  getJovoUsers(): unknown;
}
```
3. An abstract method also in `DbPlugin` class: `abstract getInstance(): DbInstance;`

Do you prefer something like this?

jrglg: Hi, @aswetlow 

Please, could you provide some advice in order to be able to get the DB client from the plugin in runtime?
I'm very new to TypeScript, and I found many difficulties on trying to do so.

The solution I raised would require DynamoDb class to implement the methods to return the client, I didn't see any problem there. But with FileDb a null would have to be returned, I think.

The most difficult part for me is changing the framework code to test it in my project. I tried for hours yesterday, but couldn't make it.

Thank you for your fantastic work, as always.

rubenaeg:
Hey @jrglg, as I wrote the MongoDb plugin originally in `v3` of the Jovo Framework,  I would like to assist you in getting this done. I'm currently in the process of reviewing your solution, let me get back to you once I have a good understanding of your code.

rubenaeg:
Looks good to me, and it also looks like it conforms to the convention to only initialize the client once. I would make some small adjustments, just minor things: 

```typescript
async initialize(): Promise<void> {
  if ((await this.jovoDb()).databaseName === this.MONGODB_DEFAULT_DATABASE_NAME) {
    // eslint-disable-next-line no-console
    console.warn('[MongoDB] Warning: The "test" database is being used.');
  }
}
```

Here you use `this.MONGODB_DEFAULT_DATABASE_NAME`, although it's a static property. I would personally go the `export const ... ` way, because this has been the convention in the framework, but if you want to keep it static, use `MongoDb.MONGODB_...` instead. Also, there is `MONGODB_DEFAULT_DATABASE_NAME` and `JOVO_DEFAULT_DATABASE_NAME`, I think would should suffice.

Please also update the docs to use `new MongoDb()` instead of `MongoDb.instance()`.

Also, in the docs you have this handler: 

```typescript
async START() {
  // You can read data from another user
  const users = mongoDb.jovoUsers();
  const otherUserData = (await users).find({ _id: '<another_id>' });
  // ...
```

Where is `mongoDb` coming from?  

Apart from that I would like to dig a bit deeper into `this.$db.client`, I like the idea to be able to store data in other collections/databases in handlers, but I think this needs more discussion, as abstracting the client across multiple database plugins is not something done easily, because you have different client implementaitons (no client for `FileDb`, client object for `DynamoDb`, client promise for `MongoDb`), schemas and architectures.
I would suggest merging the integration as is and starting a thread on the Community Forum to discuss this further.

jrglg:
hey @rubenaeg 

I appreciate the suggestions and I changed what you mentioned. 

`Where is mongoDb coming from?`
Once it was a Singleton, you pass it to the application:
```typescript
import { MongoDb } from '@jovotech/db-mongodb';
// ...
app.configure({
  plugins: [
    MongoDb.instance({
      // Configuration
    }),
    // ...
  ],
});
```
And later you import it again from a component/wherever you want.

I preffer to wait until `this.$db.client` is available and meanwhile I can use a workaround to get the client.

I've tested the package in my local project and in Lambda and it works nicely.

Thank you!

integrations/db-mongodb/README.md
jrglg: Hi Jan, I think we can resolve this as we continued with another aproach down in the comments.

rubenaeg:
@aswetlow @jankoenig what's the status on this? If you need anything else to finish this up I would be happy to help :)

jankoenig:
I'm away this week, but I'll take a look at the documentation next week. Thanks for your help!

integrations/db-mongodb/src/MongoDb.ts
```diff
@@ -0,0 +1,103 @@
...
+    await users.updateOne(filter, { $set: item }, { upsert: true });
+  }
+
+  /** MongoDB creates the database if doesn't exist yet */
+  async jovoDb(): Promise<Db> {
+    const connection = await this.client;
+    if (this.config.databaseName) {
+      return connection!.db(this.config.databaseName);
+    } else {
+      //If not provided, use database name from connection string.
```
jankoenig: I'm not sure if this is ever going to be entered because of `getDefaultConfig`

integrations/db-mongodb/src/MongoDb.ts
```diff
@@ -0,0 +1,103 @@
...
+  async saveData(userId: string, jovo: Jovo): Promise<void> {
+    const users = await this.jovoUsers();
+    const item: DbItem = { _id: userId };
+    await this.applyPersistableData(jovo, item);
+    const filter = { _id: userId };
+    await users.updateOne(filter, { $set: item }, { upsert: true });
+  }
+
+  /** MongoDB creates the database if doesn't exist yet */
+  async jovoDb(): Promise<Db> {
```
jankoenig: Should we rename this to `getDb`?

integrations/db-mongodb/src/MongoDb.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (this.config.databaseName) {
+      return connection!.db(this.config.databaseName);
+    } else {
+      //If not provided, use database name from connection string.
+      return connection!.db();
+    }
+  }
+
+  /** MongoDB creates the collection if doesn't exist yet  */
+  async jovoUsers(): Promise<Collection<Document>> {
```
jankoenig: Should we rename this to `getConnection`?

integrations/db-mongodb/src/MongoDb.ts
```diff
@@ -0,0 +1,103 @@
...
+  user?: PersistableUserData;
+  session?: PersistableSessionData;
+  createdAt?: string;
+  updatedAt?: string;
+}
+
+/** Default database name in MongoDB. if no name was specified, 'test' is used. See https://docs.mongodb.com/manual/tutorial/getting-started/#getting-started */
+export const MONGODB_DEFAULT_DATABASE_NAME = 'test';
+export const JOVO_DEFAULT_DATABASE_NAME = 'jovo_db';
+export const JOVO_DEFAULT_COLLECTION_NAME = 'jovoUsers';
```
jankoenig, line 35: Is there a specific reason why `JOVO_DEFAULT_DATABASE_NAME` uses `jovo_db` (snake_case) and  `JOVO_DEFAULT_COLLECTION_NAME` uses `jovoUsers` (camelCase)? Does this have something to do with MongoDB naming conventions?

jankoenig (changes requested):

jankoenig:
Hi @jrglg, thank you for the updates. And thanks for the feedback @rubenaeg 👍🏻 

I added a few more comments that are mainly about consistency and naming. I also updated the documentation.

There is one main thing that I'd still like to see changed before we can merge this. We need to find a way to make this happen:

```typescript
import { MongoDb } from '@jovotech/db-mongodb';
// ...

app.configure({
  plugins: [
    new MongoDb({
      // Configuration
    }),
    // ...
  ],
});
```

To access the current connection, is it possible for us to find a way to access e.g. a `this.$mongoDb` instance from the handler @aswetlow? Then we could do something like this:

```typescript
await this.$mongoDb.getConnection()
```

integrations/db-mongodb/src/MongoDb.ts
jrglg: Ok!

integrations/db-mongodb/src/MongoDb.ts
jrglg: Because the examples in MongoDb Atlas were using snake case for DB and camel case for collections. 
Seems like camel case is the convention.

integrations/db-mongodb/src/MongoDb.ts
jrglg: Ok!

integrations/db-mongodb/src/MongoDb.ts
jrglg: You are right. It's not entering there.

That's an ordinary check I put there before adding default database name in getDefaultConfig.

Since the database name can be specified either in the connection URL or in the MongoDb integration constructor (`databaseName`), I'm not sure what to do here.

If a user specify the DB name in the connection url and default config for DB name `databaseName` is present, it has no efect and will never connect to DB specified in connection URL. That's a problem I guess.

With the new client for MongoDb for Node, I can't find an elegant way to check either if DB name was specified in URL or not.

I would remove `databaseName` from `getDefaultConfig()` and set it as optional in the constructor, and leave the default collectionName like it is now. This way, a user can set the DB name either in the URL or in the constructor. By doing that, the check you referred would make sense.

Another option is to not allow a user to specify a dataBase name in the URL, but in the `databaseName` argument in the constructor. That would have to be clear in the docs to avoid future issues.

jrglg:
Hi @jankoenig,

about implementing a way to reuse connection:

**I tried to do so, but coudn't find an elegant solution**. I like what @rubenaeg said in a previous post.

> Where is `mongoDb` coming from?
> 
> Apart from that I would like to dig a bit deeper into `this.$db.client`, I like the idea to be able to store data in other collections/databases in handlers, but I think this needs more discussion, as abstracting the client across multiple database plugins is not something done easily, because you have different client implementaitons (no client for `FileDb`, client object for `DynamoDb`, client promise for `MongoDb`), schemas and architectures. I would suggest merging the integration as is and starting a thread on the Community Forum to discuss this further.

`this.db` seems better to me, to avoid having multiple DB properties under `this`.
It should provide access to all DB clients (because two can be added simultaneously), and taking into account datatypes Ruben said:

> no client for FileDb, client object for DynamoDb, client promise for MongoDb

`const client: Promise<MongoClient> = await this.$db.mongoDb.getClient()`

`const client: DynamoDBClient = this.$db.dynamoDb.getClient()`

etc.

I would be very happy with any solution that allows reusing the client, **regardless of how it is achieved**. Modifying the framework beyond an integration package is more difficult to me.

integrations/db-mongodb/src/MongoDb.ts
jrglg: Now the behaviour is:

- `databaseName` argument in MongoDb constructor takes precedence over database name in connection URL.
- If no `databaseName` is specified, it will connect to database specified in URL.
- If no database is specified in the URL nor in the databaseName, `test` database is used (deault MongoDb behaviour).

So:

1.
```
    new MongoDb({
      connectionString: 'mongodb+srv://userX@clusterX.mongodb.net/mydb',
      databaseName: 'mydb2'
    }),
``` 
`mydb2` will be used.


2.
```
    new MongoDb({
      connectionString: 'mongodb+srv://userX@clusterX.mongodb.net/mydb',
    }),
``` 
`mydb` will be used.


3.
```
    new MongoDb({
      connectionString: 'mongodb+srv://userX@clusterX.mongodb.net/',
    }),
``` 
Shows a warning in app instantiation:
`[MongoDB] Warning: The "test" database is being used.`


Note that collection name remains the same, as it can't be specified in URL.

Please tell me what you think @jankoenig, I will change it again if needed.

jankoenig: I like the new solution! Thank you 🎉

jankoenig (approved):
Thank you @jrglg for your updates and thoughts!

We implemented the `this.$mongoDb` object with offers 3 methods:

```typescript
await this.$mongoDb.getClient();
await this.$mongoDb.getDatabase();
await this.$mongoDb.getConnection();
```

I also updated the docs accordingly

jrglg:
Nice!

Thank you.

aswetlow (approved):

PR #1246, opened 2022-02-22
:bug: added missing parameter for recursive getImportStatus call

sadlowskij:
## Proposed changes
This fixes a bug, that happened when deploying alexa on an askProfile that is not the default. While the first getImportStatus call did use the correct askProfile, the recursive call if the status is 'IN_PROGRESS' dropped it, leading to an auth Error. This leads to the developer thinking that the deploy went wrong, even though it could have been deployed without issues.

## Types of changes

## Checklist


sadlowskij:
Sorry, I messed something up here, I closed this and created a clean PR

PR #1258, opened 2022-03-08
Error handling for this.repeat() if context is missing.

dominik-meissner:
Before my change a empty response was given (that might result in an platform error).

## Types of changes

## Checklist


aswetlow (changes requested):
Thank you for the fix!

Could you change `throw new Error('Context saving has to be set in config to use repeat().');`
to `throw new JovoError('Context saving has to be set in config to use repeat().', ErrorCode.ERR, 'jovo-framework');`

jankoenig (approved):

aswetlow (approved):

PR #1271, opened 2022-03-30
:sparkles: Access nlu properties in CoreRequest

rubenaeg:
## Proposed Changes

This PR modifies the `JovoInputBuilder` to accept an input object, analogous to `JovoInput`. This allows inheriting request classes to easily extend the `getInput()` function, as seen in `CoreRequest`. Furthermore, `CoreRequest.prototype.getIntents()` and `CoreRequest.prototype.getEntities()` allow access to properties in `input.nlu`. 

## Types of Changes

## Checklist


rubenaeg:
Oh, I was just wondering, would it make sense to change `getEntities()` (and `getIntents()` for that matter as well) to merge the properties instead of using logical or `||`?

platforms/platform-core/src/CoreRequest.ts
```diff
@@ -30,15 +41,17 @@ export class CoreRequest extends JovoRequest {
   }
 
   getEntities(): JovoInput['entities'] {
-    return this.input?.entities;
+    return this.input?.nlu?.entities || this.input?.entities;
```
jankoenig: The root `entities` should be prioritized:

```
return this.input?.entities || this.input?.nlu?.entities;
```

jankoenig, line 44: The root `entities` should be prioritized:

```
return this.input?.entities || this.input?.nlu?.entities;
```

platforms/platform-core/src/CoreRequest.ts
```diff
@@ -18,8 +25,12 @@ export class CoreRequest extends JovoRequest {
     return this.locale;
   }
 
+  getInput(): JovoInput {
+    return new JovoInputBuilder(super.getInput()).set('nlu', this.input?.nlu).build();
+  }
+
   getIntent(): JovoInput['intent'] {
-    return this.input?.intent;
+    return this.input?.nlu?.intent || this.input?.intent;
```
jankoenig: The root `intent` should be prioritized:

```
return this.input?.intent || this.input?.nlu?.intent;
```

jankoenig, line 33: The root `intent` should be prioritized:

```
return this.input?.intent || this.input?.nlu?.intent;
```

platforms/platform-core/src/CoreRequest.ts
```diff
@@ -18,8 +25,12 @@ export class CoreRequest extends JovoRequest {
     return this.locale;
   }
 
+  getInput(): JovoInput {
+    return new JovoInputBuilder(super.getInput()).set('nlu', this.input?.nlu).build();
```
jankoenig: Could you explain a bit further what this is doing here? Thanks!

jankoenig (changes requested):

platforms/platform-core/src/CoreRequest.ts
rubenaeg: Got it!

platforms/platform-core/src/CoreRequest.ts
rubenaeg: Done

platforms/platform-core/src/CoreRequest.ts
rubenaeg: So the idea behind this was that you can easily extend  `this.$input` by allowing inheriting request classes to build upon the `JovoInputBuilder`. Here we get the input built in `JovoRequest` and add the `nlu` property, so you can access `this.$input.nlu` in the handler.

jankoenig: Thank you. I'm still not sure if I get it, to be honest. A bit confused why specifically `nlu` needs to be set in this method and why we're not setting any other property. What is a use case here?

Could the reasoning be commented or documented somewhere? i think otherwise it could happen that we don't know why this is there if we look at the code in a few months

rubenaeg: Okay so I guess this could be seen as being redundant, the original idea was to access the nlu properties in the handler, because if intents/entities were set in `input.nlu`, you had no way of accessing it with `this.$input`, but since this PR allows accessing them through `getIntent()` and `getEntities()` it can be removed.

rmtuckerphx: @rubenaeg How do we access other properties that we might add to $input.nlu?
This could be an output object with messages if the SLU/NLU provides dialog management or messages.

rubenaeg: Well I now removed the code, but you can see in the code above how you could add properties to the input object that is returned by `getInput()`. @jankoenig do you need more input on this PR? I can write some documentation if you'd like, something like "Extending the Input object", where I could also cover ambient declarations...

jankoenig (approved):

aswetlow (approved):

PR #1279, opened 2022-04-14
:technologist: throw error in DynamoDb.initialize

sadlowskij:
## Proposed Changes

I had an issue where the connection to the dynamoDb could not be set up and it was only shown on the first request coming in, where loadData got called. As the error was not output or thrown in initialize, the information gets lost there and it seems, as if everything works fine. If there is no reason (an error that should be "ignored") for not throwing here, I think it might be helpful to have that error thrown right away in initialize.

## Types of Changes

## Checklist


jankoenig (approved):
Thank you! 🎉

aswetlow (approved):

PR #1308, opened 2022-04-29
:label: added apla directive

sadlowskij:
## Proposed Changes

I wanted to use the APLA Feature for Alexa Repromts and saw that the type did not exist yet on the Alexa reprompt type. I also could not find a class for the APLA Render Document in general, which is why I added that, and added the possibility to add this both in the response itself as well as in the reprompt directives. The reprompt directives can only contain this one directive (see [docs](https://developer.amazon.com/en-US/docs/alexa/custom-skills/request-and-response-json-reference.html#reprompt-object)). I saw that these directives have corresponding classes with class-transformer decorators to transform JSON/Pojo Objects into instances. I was however not exactly sure when this transformation happens exactly. I tried to make my changes in a way that I saw in the other Directives / Response, but am not 100% sure, I have not forgotten. I tried using it with just adding a pojo directive to the reprompt, which seems to look good now. I'd be happy about feedback!

## Types of Changes

## Checklist

### DIFF
--- a/platforms/platform-alexa/src/output/models/Reprompt.ts
+++ b/platforms/platform-alexa/src/output/models/Reprompt.ts
@@ -1,9 +1,28 @@
 import { IsOptional, Type, ValidateNested } from '@jovotech/output';
+import { IsValidDirectivesArray } from '../decorators/validation/IsValidDirectivesArray';
+import { AplaRenderDocumentDirective } from './apla/AplaRenderDocumentDirective';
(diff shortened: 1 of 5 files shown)


jankoenig:
Thank you @sadlowskij! We're planning to review this tomorrow

jankoenig (approved):
Thank you! Looks good to me ✅

aswetlow (approved):

PR #1314, opened 2022-05-05
:sparkles: Add Connections.StartConnection outputs

rmtuckerphx:
## Proposed Changes

Add the following Output classes for Connections.StartConnection:

- connection://AMAZON.VerifyPerson/2
- connection://AMAZON.LinkApp/2
- connection://AMAZON.PrintPDF/1
- connection://AMAZON.PrintImage/1
- connection://AMAZON.PrintWebPage/1
- connection://AMAZON.TestStatusCode/1
- connection://AMAZON.AskForPermissionsConsent/2
- connection://AMAZON.ScheduleTaxiReservation/1
- connection://AMAZON.ScheduleFoodEstablishmentReservation/1

## Types of Changes

## Checklist


platforms/platform-alexa/src/output/templates/ConnectionAskForPermissionConsentOutput.ts
```diff
@@ -0,0 +1,68 @@
+import { BaseOutput, Output, OutputOptions, OutputTemplate, EnumLike } from '@jovotech/framework';
+
+export enum ConsentLevel {
+  Account = 'ACCOUNT',
+  Person = 'PERSON',
+}
+
+export type ConsentLevelLike = EnumLike<ConsentLevel> | string;
+
+export enum PermissionScope {
```
aswetlow: Should we move this to `platform-alexa/src/interfaces.ts` ? This part is also used in `platform-alexa/src/output/models/card/Card.ts`

aswetlow:
Thank you for the pr @rmtuckerphx

platforms/platform-alexa/src/output/templates/ConnectionAskForPermissionConsentOutput.ts
rmtuckerphx: @aswetlow 
I refactored some interfaces, types and enums to `interfaces.ts`.
I reused the `PermissionScope` in `Card.ts` while allowing Card to have a few extra values.

rmtuckerphx:
@aswetlow I had to do another reactor to models/common to support other card classes.
There is an issue with the code that I don't know enough TypeScript to fix. The `PermissionScope` values supported in the Connections delegate are a subset of those supported by Cards.

I tried to handle it this way:

```
export enum CardPermissionScope {
  ReadList = 'read::alexa:household:list',
  WriteList = 'write::alexa:household:list',
  ReadAddressFull = 'read::alexa:device:all:address',
}

export type CardPermissionScopeLike = EnumLike<PermissionScope> | EnumLike<CardPermissionScope> | string;
```

But `AskForRemindersPermissionOutput` doesn't like `CardPermissionScopeLike.ReadWriteReminders`

aswetlow:
Thank you for the changes.

Splitting `PermissionScope` into `PermissionScope` and `CardPermissionScope` would make it a breaking change.

Is it necessary to have `CardPermissionScope` ?

rmtuckerphx:
@aswetlow There are 3 more PermissionScope values that are supported in cards than are supported by Connections.

Our options are:
1. Breaking change where CardPermissionScope  has the 3 additional items
2. Use a single list of PermissionScope  values for both Cards and Connection even though 3 would be invalid for Connections.
3. Keep PermissionScope the same and have a different ConnectionPermissionScope that include duplicates from both lists.

Originally I created it closer to option 3. The current PR is option 1.

Tell me what option you want me to implement.

aswetlow:
Thank you for the explanation. It makes sense now. We discuss it internally and I'll come back to you.

jankoenig:
Thank you @rmtuckerphx! I've just added some docs for this.

Since I haven't worked with Skill Connections myself, I'm not sure if I missed something, could you take another look?

Also, according to the [Alexa docs](https://developer.amazon.com/en-US/docs/alexa/custom-skills/use-skill-connections-to-request-tasks.html#for-managed-skill-connection-with-resume_session), there is usually an `onCompletion` property that I couldn't find in the output classes I was looking at. Any thoughts on this?

Thanks!

rmtuckerphx:
@jankoenig The docs look good. 
Good catch on `onCompletion`
@aswetlow  I've added `onCompletion` to each `ConnectionXxxOutput` class.

jankoenig (approved):
Thank you @rmtuckerphx! Ready to merge from my side ✅ 

@aswetlow and I are going to talk about the interface issues tomorrow

aswetlow:
> @aswetlow There are 3 more PermissionScope values that are supported in cards than are supported by Connections.
> 
> Our options are:
> 
> 1. Breaking change where CardPermissionScope  has the 3 additional items
> 2. Use a single list of PermissionScope  values for both Cards and Connection even though 3 would be invalid for Connections.
> 3. Keep PermissionScope the same and have a different ConnectionPermissionScope that include duplicates from both lists.
> 
> Originally I created it closer to option 3. The current PR is option 1.
> 
> Tell me what option you want me to implement.

Hey @rmtuckerphx 
Sorry for the late response. We discussed it and came to the conclusion that 3. is the best option. 

In my opinion, duplicates are not always bad because they often increase comprehensibility.

jankoenig:
> Hey @rmtuckerphx Sorry for the late response. We discussed it and came to the conclusion that 3. is the best option.
> 
> In my opinion, duplicates are not always bad because they often increase comprehensibility.

I just made the according changes and fixed merge conflicts.

Ready to merge from my side ✅

aswetlow (approved):

jankoenig:
Thank you @rmtuckerphx 🎉

PR #1350, opened 2022-06-10
:adhesive_bandage: only awaiting output.build if it returns promise

sadlowskij:
## Proposed Changes

This change tries to ensure the order of outputs to be consistent with the order of calls as much as possible, by preventing awaiting `build` calls of outputs that do not return a promise.
With this Output:

```
@Output()
export class YesNoOutput extends BaseOutput {
  build(): OutputTemplate | OutputTemplate[] {
    return {
      quickReplies: ['yes', 'no'],
      listen: true,
    };
  }
}
```
this handler 

```
  START() {
    this.$send(YesNoOutput, { message: 'Do you like Pizza?' });
    return this.$send('this should come after the YesNoOutput');
  }
```
would result in the outputs being in the order that they have been called, while without this, the order would be switched, as `YesNoOutput` is awaited even if the `build` does not return a Promise. I am aware that this can be fixed by awaiting every `$send` call (or at least the ones containing an Output class), but first guess was, that you would only have to await if the `build` method was returning a Promise. I thought that this, or something like this could possibly prevent similar mistakes being made in the future. Maybe this is also just how it is supposed to be though or is intended behavior, in that case, this can just be closed! 

One possible issue I found with this approach is, if I got that right, not only Promises, but all objects with a `then` function can be awaited (see [here](https://stackoverflow.com/questions/27746304/how-do-i-tell-if-an-object-is-a-promise)) but I think this should not be an issue, as type checking ensures for the return type to be an actual Promise. If you consider adding this, I could look into this further though!

While handling this issue I was also wondering, if the actual best approach would also just to be to keep the order always the same as it has been called (even for async `build` calls). Maybe that could be done by somehow adding the Promises to the `$output` array and awaiting the non resolved Promises at the end, or just storing the Promise with the order information and add it to `$output` in the correct order when it resolves. This might have other side effects, that I might not see right now though.

Does one of these approaches makes sense to you?

## Types of Changes

## Checklist

### DIFF
--- a/framework/src/Jovo.ts
+++ b/framework/src/Jovo.ts
@@ -3,6 +3,7 @@ import { JovoResponse, NormalizedOutputTemplate, OutputTemplate } from '@jovotec
 import _cloneDeep from 'lodash.clonedeep';
 import _merge from 'lodash.merge';
 import _set from 'lodash.set';
+import util from 'util';
 import { App, AppConfig } from './App';
 import { HandleRequest } from './HandleRequest';
 import {
@@ -263,7 +264,8 @@ export abstract class Jovo<
     let newOutput: OutputTemplate | OutputTemplate[];
     if (typeof outputConstructorOrTemplateOrMessage === 'function') {
       const outputInstance = new outputConstructorOrTemplateOrMessage(this, options);
-      const output = await outputInstance.build();
+      const outputRes = outputInstance.build();
+      const output = util.types.isPromise(outputRes) ? await outputRes : outputRes;
       // overwrite reserved properties of the built object i.e. message
       NormalizedOutputTemplate.getKeys().forEach((key) => {
         if (typeof options?.[key] !== 'undefined') {


jankoenig:
Thank you @sadlowskij! We're planning to review this tomorrow

aswetlow:
Hey @sadlowskij 
Thank you for the PR. It does definitely improve the prevention of unexpected behavior.

aswetlow (approved):

jankoenig (approved):
Thank you @sadlowskij 🎉

jankoenig:
> While handling this issue I was also wondering, if the actual best approach would also just to be to keep the order always the same as it has been called (even for async `build` calls). Maybe that could be done by somehow adding the Promises to the `$output` array and awaiting the non resolved Promises at the end, or just storing the Promise with the order information and add it to `$output` in the correct order when it resolves. This might have other side effects, that I might not see right now though.

Sounds interesting! Just talked about this with @aswetlow and we think there might be some issues regarding serializability. We think it's best to merge this PR as is and then investigate other potential issues with the output order at a later point.

Thanks for your work!

sadlowskij:
Yes, it certainly is a bigger change that would have to be thought through in detail, the state with the merged PR will improve this to a state where it's certainly easier to anticipate the order! Thanks for merging!

PR #1364, opened 2022-06-28
📝 Update handle decorators docs

itaybel:
## Proposed Changes

Usually when referring to a camelCase, its more reasonable to type camelCase instead of camel case.


jankoenig (approved):
Makes sense. Thank you!

PR #1398, opened 2022-08-06
❇️ Add S3 TTS Cache

rmtuckerphx:
## Proposed Changes

Add S3 TTS Cache

## Types of Changes

## Checklist


rmtuckerphx:
@aswetlow @jankoenig  I wanted to merge just the ttscache-s3 files in this PR but accidently also included the commits for tts-polly.
Can you help me untangle this?

integrations/ttscache-s3/src/S3TtsCache.ts
```diff
@@ -0,0 +1,153 @@
...
+import { DeepPartial } from '@jovotech/common';
+
+import { Readable } from 'stream';
+
+import { urlJoin } from 'url-join-ts';
+
+import { TtsCachePlugin, TtsCachePluginConfig, TtsData, AudioUtilities } from '@jovotech/framework';
+
+export interface S3TtsCacheConfig extends TtsCachePluginConfig {
+  credentials: Credentials;
```
aswetlow: Is it also required in a Lambda runtime?

integrations/ttscache-s3/src/S3TtsCache.ts
```diff
@@ -0,0 +1,153 @@
...
+      if (response.ContentType) {
+        return {
+          key,
+          fileExtension,
+          contentType: response.ContentType,
+          url: fullPath,
+        };
+      }
+    } catch (error) {
+      console.log((error as Error).message);
```
aswetlow, line 93: What are possible errors? Should we pass the error to the caller?

aswetlow:
> @aswetlow @jankoenig I wanted to merge just the ttscache-s3 files in this PR but accidently also included the commits for tts-polly. Can you help me untangle this?

Done

integrations/ttscache-s3/src/S3TtsCache.ts
rmtuckerphx: A Lambda runtime is not required, that is why the credential values. But is there a way to check so that if you are in a Lambda you can use those permissions? Any thoughts on how to do this? Should Credentials in config be optional?

integrations/ttscache-s3/src/S3TtsCache.ts
rmtuckerphx: What are your ideas on surfacing the error. Maybe a LastError property exposed on a $ttscacheS3 object? Or an OnError callback on the Plugin class? A cache error should just behave as if no cache exists.

jankoenig: I agree that a cache error shouldn't cause the app to break

aswetlow: Yes, it shouldn't break anything.

The error delegation needs more design and concept work and is probably something that is not relevant for this PR.

integrations/ttscache-s3/src/S3TtsCache.ts
aswetlow: I was unclear, sorry. We don't need Lambda, but it should work on Lambda without adding the credentials. (like for DynamoDB)

As far as I know, all of the AWS SDKs look for the environment variables. 
Optional credentials object could be sufficient.

aswetlow: It could make more sense to remove `credentials` and add `libraryConfig` like in the DynamoDB plugin. This would make it more flexible with all the config properties.
image

integrations/ttscache-s3/src/S3TtsCache.ts
```diff
@@ -0,0 +1,153 @@
...
+import { Readable } from 'stream';
+
+import { urlJoin } from 'url-join-ts';
+
+import { TtsCachePlugin, TtsCachePluginConfig, TtsData, AudioUtilities, RequiredOnlyWhere } from '@jovotech/framework';
+
+export interface S3TtsCacheConfig extends TtsCachePluginConfig {
+  bucket: string;
+  path: string;
+  baseUrl: string;
```
jankoenig: Hey @rmtuckerphx, @aswetlow and I are finishing up the review and have one last question regarding `baseUrl`:

In your example, you have `bucket` as `mybucket-public` and `baseUrl` as `https://mybucket-public.s3.amazonaws.com`.

Does the URL always look like this format? `https://<bucket>.s3.amazonaws.com`
If yes, we could maybe remove `baseUrl` and put the string together ourselves.

What do you think?

rmtuckerphx: @jankoenig I know in the past that the url format changed for newer buckets, so I didn't want to tie these too closely.

rmtuckerphx: @jankoenig @aswetlow  I dug up this old article that explained the transition of bucket urls:
https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

Reading this again, I think that going forward we only need to support virtual-hosted style URL so we can remove `baseUrl` from config and construct it in code.

Thanks for adding `libraryConfig` to support `S3ClientConfig`. That is the better approach.

jankoenig (approved):
Thank you @rmtuckerphx ✅

aswetlow (approved):

Issue #1251, opened 2022-02-24
Alexa Lists in v4


NLueg:
## I'm submitting a...

- [ ] Bug report
- [ ] Feature request
- [x] Documentation issue or request
- [ ] Other... Please describe:

## Expected Behavior

It should be documented how to ask for permissions to lists and how to add items to the users list.
I actually found a quite old fork where it's documented: https://github.com/gwythyr/jovo-framework-nodejs/blob/master/docs/05_platform-specifics/amazon-alexa/lists.md

## Current Behavior

I have no idea how to include the users list.
I don't found any documentation about this.

## Your Environment

* Jovo Framework version used: 4.2.2

jankoenig:
We don't have convenience classes for this yet, but the permission cards could be added to the response using the [`nativeResponse` property](https://www.jovo.tech/marketplace/platform-alexa/output#native-response).

It might also be helpful to look at the reminders documentation, because the implementation would probably look similar: https://www.jovo.tech/marketplace/platform-alexa/reminders#permissions-consent-card

Happy to take a look at a PR if anyone from the community wants to build this

NLueg:
@jankoenig Thank you for your answer. I will try to implement this by myself and will create a PR for this.
I currently struggle with the Events that Alexa can send to a skill. Not sure, if this is a new issue, but for now I will leave it in a comment.

In Jovo v3 there was an `ON_EVENT` handler (for example like here: https://github.com/jovotech/jovo-templates/blob/7939c654191bc961ecec9b91eac27c29295043e5/alexa/skillevents/typescript/src/app.ts)
I don't found any similar concept for Jovo v4.

aswetlow:
Hey @NLueg 

The `@Handle` decorator is a powerful tool for these use cases. https://www.jovo.tech/docs/handle-decorators 

```typescript
@Handle({
  global: true,
  types: ['AlexaSkillEvent.SkillDisabled'],
  platforms: ['alexa'],
})
handleSkillDisabledEvent() {
  // 
}
```

NLueg:
Ah, okay perfect. I actually looked into the docs, but I think I'm blind 😄 Thank you very much for the fast response!

NLueg:
I actually managed it to implement the functionality in Jovo v4.
Besides the new Handlers I oriented myself heavily on the reminder implementation of the framework.
[...]

PR #1418, opened 2022-09-07
✨ Add Alexa Lists feature

NLueg:
## Proposed Changes

As requested in #1251 here is the way I managed it to include the Alexa lists api with Jovo.
I've added a new documentation section for it and added an example which shows how to include it in the real world.

This PR also contains some changes to the api to make it easier to get an item from a specific list.
Lastly, it contains a small bugfix as the current permission string for Alexa lists is wrong.

I still have an issue with uploading the package, but I found a workaround for it.
I don't include the workaround here, but I could think about doing this. You can find more information about it in #1419 

Can't wait for your feedback!

## Types of Changes

## Checklist

### DIFF
--- a/examples/typescript/alexa/lists/src/components/GlobalComponent.ts
+++ b/examples/typescript/alexa/lists/src/components/GlobalComponent.ts
@@ -1,4 +1,4 @@
-import { BaseComponent, Component, Global, Handle, Jovo } from '@jovotech/framework';
+import { BaseComponent, Component, Global, Handle } from '@jovotech/framework';
 import { LoveHatePizzaComponent } from './LoveHatePizzaComponent';
 
 @Global()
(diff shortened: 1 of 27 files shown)


aswetlow (approved):

jankoenig:
Thanks a lot @NLueg 🎉 

I'll take a look at the docs early next week

examples/typescript/alexa/lists/src/components/GlobalComponent.ts
```diff
@@ -0,0 +1,66 @@
...
+    console.log(`Deleted ${listItemIds} from ${listId}`);
+  }
+}
+
+type ListItemRequest = {
+  listId?: string;
+  listItemIds?: string[];
+};
+
+function getIdsFromRequest(jovo: Jovo): ListItemRequest {
```
jankoenig: Thanks for your PR @NLueg! Looks great.

Just one question about this `getIdsFromRequest` helper: Is this maybe something that could be added as a method, either to the `ListApi` or the `AlexaRequest` class? Ideally, I'd like to keep handlers and utility functions as separate as possible.

NLueg: Jep, this totally makes sense. Moved the function to the `ListApi`!

jankoenig: Thank you @NLueg! After going through the Alexa docs and the code a bit more and discussing it with @aswetlow, I decided to take out the helper method because it's one line of code in the handler. Apologies for the stress!

jankoenig (approved):
Thanks a lot @NLueg 🎉

aswetlow (approved):

Issue #1211, opened 2022-02-01
`Unhandled` and `@PrioritizedOverUnhandled` handlers don't clean up the state stack correctly


jtfell:
## I'm submitting a...

- [x] Bug report
- [ ] Feature request
- [ ] Documentation issue or request
- [ ] Other... Please describe:

## Expected Behavior

I'll use `@PrioritizedOverUnhandled` as the example here, but I believe this issue applies more generally when delegated components aren't explicitly resolved.

If I have a handler with the `@PrioritizedOverUnhandled()` decorator for a specific intent and its called, the state stack is left in a way that breaks the `$resolve` behaviour.

See the reproduction [here](https://github.com/jtfell/jovo-repro).

## Current Behavior

I did this interaction through the debugger:

![image](https://user-images.githubusercontent.com/6702746/151897110-b587c14b-4afa-4b13-869d-35c060ff8b6a.png)

And saw this in the state stack. It did not find the handler in the `GlobalComponent` via the `resolve` call in `LoveHatePizzaComponent`

```json
 >>>>> Request - 2022-02-01T00:53:32.267Z
{
  "version": "4.0",
  "platform": "jovo-debugger",
  "id": "d79b0a70-3c4d-4efd-947e-6cae966b51cc",
  "timestamp": "2022-02-01T00:53:31.979Z",
  "timeZone": "Australia/Brisbane",
  "locale": "en",
  "input": {
    "intent": "NoGoodIntent",
    "type": "INTENT"
  },
  "context": {
    "device": {
      "id": "b019fe87-cae7-4f79-a5cb-fb6e38e0fe29",
      "capabilities": [
        "SCREEN",
        "AUDIO"
      ]
    },
    "session": {
      "id": "0bf34a99-531a-4969-b694-e26eaaf6849f",
      "data": {},
      "state": [
        {
          "component": "GlobalComponent"
        },
        {
          "resolve": {
            "fail": "Fail"
          },
          "component": "LoveHatePizzaComponent"
        },
        {
          "resolve": {
            "yes": "lovesPizza"
          },
          "config": {
            "outputOpts": {
              "message": "Do you like pizza?"
            }
          },
          "component": "LoveHatePizzaComponent.YesNoComponent",
          "data": {
            "numberOfTries": 0
          }
        }
      ],
      "isNew": false,
      "updatedAt": "2022-02-01T00:53:30.065Z",
      "new": false
    },
    "user": {
[...]

PR #1438, opened 2022-10-13
🐛 Analyze all $state elements for local route matches

### LINKED ISSUE #1380: @PrioritizedOverUnhandled() on not global handler gets ignored
## I'm submitting a...

- [x] Bug report
- [ ] Feature request

## Expected Behavior

The [docs](https://www.jovo.tech/docs/routing#request-routing) says that handlers from all the components that are part of the $state stack and global handlers from all root components are selected for the matches array.
Now, I've ComponentA which has a global handler and a "local" one (both of them have @PrioritizedOverUnhandled() annotation) and ComponentB which has a local handler and UNHANDLED. Let's assume that the current $state stack is [ComponentA, ComponentB], so now I'm inside ComponentB.
If I invoke the intent handled by the global handler in ComponentA, the handler is correctly triggered. The problem is that if I invoke the other intent, the "local" handler isn't triggered and the request gets handled by the UNHANDLED in ComponentB.
From the docs, I assume that the local handler in ComponentA should be in the matches array as well, since ComponentA is in the current $state stack. But apparently, even if the local handler has @PrioritizedOverUnhandled() annotation, it isn't considered by the router.

## Current Behavior

You can see current behaviour in this repo [vocalime-base/jovo-4-router-issue](https://github.com/vocalime-base/jovo-4-router-issue).

Just run `npm test`.

## Your Environment

- Jovo Framework version used: 4.2.20

palle-k:
## Proposed Changes

This PR fixes #1380.
Previously, during route matching, only the topmost stack entry was analyzed for local matches. 
However, the documentation mentions that every stack entry should be matchable. 

When a component other than the topmost component of the $state stack is matched, the $state stack is unwound to that component.

**Note**: While this makes the behavior of Jovo consistent with the documentation, this is technically a breaking change.

## Types of Changes

- [x] Bug fix

## Checklist


palle-k:
@aswetlow please review.

Also: While this is out of scope for this PR, I am interested how global handlers (`@Handle({global: true, ...})`, **not** global components) fit into this picture.

When a global handler is matched that does not belong to any component of the $state stack, should the entire stack be replaced? Currently, global handlers just add components to the state stack.

acerbisgianluca:
> When a global handler is matched that does not belong to any component of the $state stack, should the entire stack be replaced? Currently, global handlers just add components to the state stack.

I would like to leave my point of view about this. I think that Jovo should replace the entire stack in that case, because the user basically jumps out from the current state's scope and should restart from a new stack.

palle-k:
This PR also potentially resolves #1211.

aswetlow (approved):

jankoenig (approved):
Thanks a lot @palle-k! 🎉

jankoenig:
> > When a global handler is matched that does not belong to any component of the $state stack, should the entire stack be replaced? Currently, global handlers just add components to the state stack.
> 
> I would like to leave my point of view about this. I think that Jovo should replace the entire stack in that case, because the user basically jumps out from the current state's scope and should restart from a new stack.

This is something I've been thinking about for a while and it's difficult for me to reach a good decision here.

- I understand the idea of deleting the stack because some different path has been entered
- However, I'd love to be able to use global handlers as "detours" as well: e.g. you're in a checkout flow and then ask something like "what is the shipping cost again?", it answers that and then leads you back to the checkout flow

Nonetheless, we need a good way of cleaning up the state stack if a global handler is reached. Would love to discuss this more, maybe in a separate issue

Thanks for the input!

palle-k:
> However, I'd love to be able to use global handlers as "detours" as well: e.g. you're in a checkout flow and then ask something like "what is the shipping cost again?", it answers that and then leads you back to the checkout flow

Could the `$resolve` method be adapted for this, so that there is a default resolution path that does not require any contract between the original component and the component with the global handler? 

This could also be interesting for delegation in general. Some components that I am writing only have a single resolution path, so having a default resolution here could be a nice to have as well. 

There could be a new special method name `CONTINUE` (like `START`/`UNHANDLED`), that would (if present) be executed upon default resolution.

PR #1445, opened 2022-11-04
:sparkles: check component availability

palle-k:
## Proposed Changes

This PR checks component availability when executing a handler.
While there has been a `isAvailable` property declared in `@Component`, it has never been checked.

## Types of Changes

## Checklist

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -220,6 +220,7 @@ For some components, it may be helpful (or necessary) to add options for customi
 - `config`: The custom config used by the component. Can be accessed with `this.$component.config`.
 - `models`: Model files for component-specific intents and entities (_in development_).
 - `name`: If two components have the same class name, one component's name can be changed here.
+- `isAvailable`: A function that returns a boolean whether the component is available. If it returns `false`, the component is skipped during routing and redirecting or delegating to it causes a `ComponentNotAvailableError`.
 
 In the [register root components](#register-root-components) section, we already talked about how to pass options when registering existing components.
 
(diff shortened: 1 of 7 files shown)


jankoenig:
Thank you @palle-k!

I have 2 questions:
- Could you share a use case how you'd like to use this feature? This would help us evaluate the design/write documentation etc.
- You're throwing an error when a component isn't available. In your Jovo app, what would you do if the component weren't available and it threw an error, would you catch the error somewhere and then redirect to a different component? I'm thinking if there is maybe a different way to do that, e.g. don't take unavailable components into account during routing.

Appreciate your work!
Thanks,
Jan

palle-k:
> Could you share a use case how you'd like to use this feature? This would help us evaluate the design/write documentation etc.

- One example could be that certain components are only available when the user has a valid access token. That way, everywhere in the component, we can guarantee that the user is actually logged in. 
- Another example could be that a component is only available if certain config options are set correctly. 
- Lastly, certain components may only be available when the device supports certain features.

So in general, I would say that the `isAvailable` function should be used to check invariants based on the current component config, the session data as well as user and device data.

Also, the primary reason why I made this PR is because I wanted to ensure that the `isAvailable` property, that has already been declared, is implemented and works as expected.

> You're throwing an error when a component isn't available. In your Jovo app, what would you do if the component weren't available and it threw an error, would you catch the error somewhere and then redirect to a different component? I'm thinking if there is maybe a different way to do that, e.g. don't take unavailable components into account during routing.

That's a good point. Handling it during routing sounds more elegant. That's also the reason why this PR is still a draft. 
If we do not take unavailable components into account during routing, what would you expect to happen when a redirect or delegation towards such a component occurs? That should still result in an error, right? 

In my Jovo app, I built a system of error boundaries (like error boundaries in React), where each component can have a `ERROR` method, which would catch the error and resume execution. The nearest error boundary would then also be responsible for catching this error. If you are interested, I could share the error boundary implementation as well.

palle-k:
I have updated this PR to take `isAvailable` into account in the `RoutingExecutor`.

jankoenig (approved):
Looks great. Thank you! 🎉

aswetlow (approved):

PR #1463, opened 2022-11-12
:sparkle: Add TextReplaceNluPlugin

rmtuckerphx:
## Proposed Changes

Add TextReplaceNluPlugin
Replace input text for all matches of a search value (string or RegExp) based on rules as an NLU preprocessor.

## Types of Changes

## Checklist

### DIFF
--- a/integrations/plugin-textreplacenlu/.npmignore
+++ b/integrations/plugin-textreplacenlu/.npmignore
@@ -1,0 +1,5 @@
+.idea
+node_modules
+npm-debug.log
+/src
+/test
--- a/integrations/plugin-textreplacenlu/README.md
+++ b/integrations/plugin-textreplacenlu/README.md
@@ -1,0 +1,20 @@
+# Jovo Text Replace NLU Plugin
+
+[![Jovo Framework](https://www.jovo.tech/img/github-header.png)](https://www.jovo.tech)
+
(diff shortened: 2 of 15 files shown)


rmtuckerphx:
@jankoenig @aswetlow This is a rework of a previous PR that you asked be its own plugin.
I ended up modifying the base `tsconfig.json` by adding `ES2021.String` so I could access `string.replaceAll()`. Could use some help figuring out the error I'm getting.

aswetlow:
Hey Mark, thank you for the PR. `replaceAll` is very convenient, but wouldn't `str.replace(/regex/g)` fix the problem?

rmtuckerphx:
@aswetlow - If search argument is a string, replaceAll() replaces all occurrences of search with replaceWith, while replace() only the first occurrence

I intended that string replacement would be the primary use case and added RegExp since it was supported by replaceAll.

Can you explain the reason behind the multiple tsconfig.build files and why replaceAll is failing?

rmtuckerphx:
@aswetlow Fix this however you want. Either option works for me.

aswetlow:
Sorry for the late response: 

Replacing `replaceAll`with `replace` should do the trick since you pass a `TextExp` object with a `g` flag.

PR #1406, opened 2022-08-22
❇️  Add event.$t middleware

rmtuckerphx:
## Proposed Changes

Add event.$t middleware to provide an easy way to listen to and track analytics on translation keys.

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [x] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1478, opened 2022-12-10
:sparkles: Add Google BigQuery Analytics plugin

rmtuckerphx:
## Proposed Changes

Track analytics using Google BigQuery.

## Types of Changes

## Checklist


integrations/analytics-bigquery/docs/README.md
```diff
@@ -0,0 +1,276 @@
...
+### Custom Events
+
+Include your own custom events by calling `$bigquery.addEvent(event)` or `$bigquery.addError(error)`.
+
+For any custom properties you add in code, update the Google BigQuery table schema to include them.
+
+#### Event
+
+```typescript
+await this.$bigquery.addEvent({
```
aswetlow: I think we should change `$bigquery` to `$bigQuery` since the official name is Google BigQuery.

integrations/analytics-bigquery/src/BigQueryAnalytics.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  async trackResponse(jovo: Jovo): Promise<void> {
+    await jovo.$bigquery.processResponse(jovo);
+  }
+
+  async eventExecuteHandler(jovo: Jovo, payload: AnyObject | undefined): Promise<void> {
+    await jovo.$bigquery.executeHandler(jovo, payload);
+  }
+
+  // async eventT(jovo: Jovo, payload: AnyObject | undefined): Promise<void> {
```
aswetlow: Please remove unused code, thank you.

integrations/analytics-bigquery/src/JovoBigQuery.ts
```diff
@@ -0,0 +1,313 @@
...
+export class JovoBigQuery {
+  readonly client: BigQuery;
+  private onError?: (error: Error) => Promise<void>;
+  private events: AnalyticsEvent[] = [];
+  readonly requestId: string;
+  readonly requestStart: number;
+
+  constructor(readonly plugin: BigQueryAnalytics, readonly jovo: Jovo) {
+    this.client = new BigQuery(this.config.libraryOptions);
+    this.requestId = uuidv4();
```
aswetlow: This makes me think of abstracting `requestId` and adding it to the  `$jovo` instance. There are many use cases. 

We also use a `requestId` in our Jovo Inbox plugin.

aswetlow: Using the Jovo Inbox and the BigQuery plugin at the same time would mean there's the same `requestId` for the same request. Debugging is a lot easier with consistent data. 

Does it make it sense?

integrations/analytics-bigquery/src/JovoBigQuery.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    if (this.config.onAddEvent) {
+      await this.config.onAddEvent(this.jovo, event);
+    }
+
+    this.events.push(event);
+
+    if ((this.config.logging as BigQueryLoggingConfig).addEvent) {
+      console.log('BigQuery addEvent', { event });
```
aswetlow, line 102: Could you add this above every line you use `console` ?
`// eslint-disable-next-line no-console`

integrations/analytics-bigquery/src/JovoBigQuery.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+  }
+
+  async addError(error: Error) {
```
aswetlow: The linter requests a return type :)

integrations/analytics-bigquery/src/JovoBigQuery.ts
```diff
@@ -0,0 +1,313 @@
...
+  get config(): BigQueryAnalyticsPluginConfig {
+    return this.plugin.config;
+  }
+
+  async processRequest(jovo: Jovo): Promise<void> {
+    this.initOnError(jovo);
+
+    if (jovo.$session.isNew) {
+      this.jovo.$session.data.bigquery = {
+        sessionStart: Date.now(),
```
aswetlow: I think this is not necessary. You can use `jovo.$session.createdAt`

aswetlow (changes requested):
🚀 Thank you for the PR @rmtuckerphx 
I added a few comments and requested changes.

integrations/analytics-bigquery/src/JovoBigQuery.ts
rmtuckerphx: @aswetlow  That makes sense, are you going to refactor `requestId`?

integrations/analytics-bigquery/src/BigQueryAnalytics.ts
rmtuckerphx: @aswetlow  I'll remove. But I left that there so you could consider it in context of my other PR https://github.com/jovotech/jovo-framework/pull/1406

rmtuckerphx:
@aswetlow I think this is ready to go

jankoenig:
Hi @rmtuckerphx,

Thank you for the PR! Great work.

I just updated the docs and everything is good to go except one thing I'd like to talk to you about. The service account key file and `libraryConfig`.

Instead of providing the `projectId` and the `fileName` (or path, which I think is a bit unintuitive as we already import the file) as part of the `libraryConfig`, would it make sense to pass the whole `serviceAccount` file? We do the same thing in Google`s Business Messages platform integration: https://www.jovo.tech/marketplace/platform-googlebusiness#configuration

Do you think this would work? We could still use `libraryConfig` for additional, optional client configurations.

Thanks,
Jan

rmtuckerphx:
@jankoenig I'm open to the changes for `libraryConfig` if they work. Not that familiar with Google.

sadlowskij:
Hey @jankoenig and @rmtuckerphx! I took a look into this, experimented a bit and found this looked at the documentation of the  [BigQueryOptions](https://googleapis.dev/nodejs/bigquery/latest/global.html#BigQueryOptions), which shows that there are multiple ways of authenticating, including using the serviceaccount json.

Here are my propositions to go further with this

1. Remove throwing erros when `keyFilename` or `projectId` are not provided in the `libraryOptions` as these are not required (see docs). This way it is up to the user how they want to authenticate themselves, e. g. by using the `GCLOUD_PROJECT` enivornment variable.
2. Add example in the docs how the serviceAccount is used like it is in used in Google Business Messages. In this case the serviceAccount is used like this
```
new BigQueryAnalytics({
  appId: 'jovo',
  datasetId: 'jovo',
  tableId: 'jovo',
  libraryOptions: {
    credentials: serviceAccount,
  },
}),
```
The one thing that is confusing to me is that the credentials interface in the BigQueryOptions does not include `projectId` and the other fields but only `client_email` and `private_key`, but just providing the whole imported serviceAccount JSON with with `projectId` and all works just fine in my tests.

Doing it this way makes it very flexible for the user, when it comes to authentication. The only thing that's still unclear is why `credentials` in the `BigQueryOptions` has this restricted interface.

If this approach is okay with you, I can make a PR to Marks Fork.

jankoenig:
Thank you @sadlowskij, this sounds great!

sadlowskij:
Okay this is the [PR](https://github.com/rmtuckerphx/jovo-framework/pull/1)

rmtuckerphx:
@aswetlow I've lost track of what is blocking this now. Please advise.

jankoenig (approved):

PR #1541, opened 2023-04-21
:bug: fixed js require in NlpjsNlu

sadlowskij:
## Proposed Changes

When I change my lang model to a javascript file (in order to benefit from type definitions and comment function) the JovoDebugger (more concretely NlpjsNlu) throws an Error `Cannot find module 'models/en.js'`. 

When using `.json` this works well, as this is imported with `fs.readFile(filePath)` which uses the current working directory for relative paths:

> Relative paths will be resolved relative to the current working directory as determined by calling process.cwd().

see [docs](https://nodejs.org/api/fs.html#fs_string_paths)

`require` does not do that out of the box, which is why I added this in that case.

## Types of Changes

## Checklist


aswetlow (approved):

aswetlow:
Thank you!

jankoenig (approved):

PR #1550, opened 2023-05-01
📄 update README

ShubhamChoudharyShubh:
📄 update README file content


jankoenig:
Thank you for the PR, but we don't think these changes in wording are necessary

PR #1562, opened 2023-06-28
:bug: Removed source-map-support for ts-jest.

### LINKED ISSUE #1561: [BUG] Initializing the TestSuite Class in the test file, create a wrong sourcemap to jest. (Show the wrong error line in the error log)
## I'm submitting a...

- [X] Bug report
- [ ] Feature request

## Expected Behavior

Installing the latest version of Jovo framework the test works as expected. 

## Current Behavior

Changing the sample.test.ts with this code: 
```typescript

import { TestSuite, InputType } from "@jovotech/framework";

const testSuite = new TestSuite();
test("new test", async () => {

  expect(1).toEqual(2);
});

```

Running the test with `npm run test` the line of the error showed is shifted by some line above:

image

This does not happen after commenting the `const testSuite = new TestSuite();`.

## Your Environment

Node: v16/v18/v19/v20

- Jovo Framework version used: 4.5.9
- Operating System: MacOS/Linux/Github CodeSpace

jBernavaPrah:
## Proposed Changes

Fix: https://github.com/jovotech/jovo-framework/issues/1561

## Types of Changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [?] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/framework/src/index.ts
+++ b/framework/src/index.ts
@@ -1,8 +1,11 @@
 import { JovoLogger } from '@jovotech/common';
 import axios from 'axios';
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-//require('source-map-support').install();
+// do not use source map support with jest.
+if (process.env.JEST_WORKER_ID === undefined) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  require('source-map-support').install();
+}
 
 export const Logger = new JovoLogger();
 
--- a/framework/src/index.ts
+++ b/framework/src/index.ts
@@ -2,7 +2,7 @@ import { JovoLogger } from '@jovotech/common';
 import axios from 'axios';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-require('source-map-support').install();
+//require('source-map-support').install();
 
 export const Logger = new JovoLogger();
 


jBernavaPrah:
Hi! 
This is my first PR here. If I did something wrong let me know :)  

Thanks!

aswetlow:
Hey @jBernavaPrah 
Great catch! I had the same issue a few weeks ago but couldn't find a quick fix. I just realized we had a solution for this in v3 https://github.com/jovotech/jovo-framework/blob/56a8ae4df80061e4d30c3bc6d91b93774cef7a1e/jovo-core/src/index.ts#L15

Removing the install part would break stuff. But it will work if it is not used in Jest. Could you update your code?

```typescript
try {
  // do not use source map support with jest.
  if (process.env.JEST_WORKER_ID === undefined) {
    // eslint-disable-next-line @typescript-eslint/no-var-requires
    require('source-map-support').install(); 
  }
} catch (error) {
  Log.error(error);
}

```

jBernavaPrah:
Thanks @aswetlow! 

I will do the change you suggested.

I have a concern regarding the try-catch.
I think that is wrong the wrap the code in the try-catch because in any case, as being impossible to install the source-map-support the process needs to fail.

Mainly 3 reasons:
1) You tell that this is necessary for another part of the project.
2) Until now, this was already without the try-catch and “worked”.
3) With only console error, this will not be possible to catch the error in upper statements.

What do you think or what I'm missing here?

aswetlow:
Sorry, I did a simple copy&paste from the old version. I can't remember why we added the try/catch block, but I think it failed in some environments. It may not be relevant anymore since we didn't have issues with it in v4.

It's a simple error logging in the catch block because it was okay to fail silently. 

You can remove the try/catch block.

```typescript
if (process.env.JEST_WORKER_ID === undefined) {
    // eslint-disable-next-line @typescript-eslint/no-var-requires
    require('source-map-support').install(); 
  }
```

Thank you for questioning my suggestion.

jBernavaPrah:
Fix done! :) 

Thanks for explaining it @aswetlow!

aswetlow (approved):

jankoenig (approved):
Thank you @jBernavaPrah! 🎉

PR #1567, opened 2023-07-03
Improve Types for $delegate and $resolve function

jBernavaPrah:
## Proposed Changes

Added a new abstract Class `BaseDelegateComponent` and an `ExtractDelegatedEventData` type to permit better use of $delegate and $resolve functions.

The `BaseDelegateComponent`, used in a similar way as the BaseComponent, is there to add the typing to events/resolve. 
The `ExtractDelegatedEventData` is used to extract the data from the event. 

Examples (pseudo code): 

```typescript
// components/DelegatedComponent.ts
class DelegatedComponent extends BaseDelegatedComponent<{
  completed: boolean, // completed is the event, boolean is the parameter, for multiple params, use array like: [boolean, string, string[]]
}> {
 async START() {
   return this.$resolve('completed', true) // <-- Typed
   
  //return this.$resolve('wrongEvent', "orWrongType")  <-- Error type from typescript
 }

}

// components/GlobalComponent
class GlobalComponent extends BaseComponent {

  async intent(){
    this.delegate(DelegatedComponent, {
    resolve: {
          completed: this.onDelegateCompleted // completed will be suggested by IDE
          // For back-compatibility, adding something wrong will NOT result in an error. We can discuss it if we think this can be useful.
    }
    })
  }
  
  onDelegateCompleted(result: ExtractDelegatedEventData<DelegatedComponent, 'completed'>){
  //  result is typed  
  if(result) return this.$send("Result frue");
  return this.$send("Result false");
  }

}
```

Completely open to discussing it in any way. 

## Types of Changes

- [] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [-] I have added tests to cover my changes

### DIFF
--- a/framework/src/BaseDelegateComponent.ts
+++ b/framework/src/BaseDelegateComponent.ts
@@ -1,5 +1,5 @@
 import { UnknownObject } from '@jovotech/common';
-import { BaseComponent, ComponentData } from './';
+import { BaseComponent, ComponentData } from './index';
 
 export type ExtractDelegatedEventData<
     T extends BaseDelegateComponent<Record<string, unknown>>,
--- a/framework/src/BaseDelegateComponent.ts
+++ b/framework/src/BaseDelegateComponent.ts
@@ -1,0 +1,30 @@
+import { UnknownObject } from '@jovotech/common';
+import { BaseComponent, ComponentData } from './';
+
+export type ExtractDelegatedEventData<
+    T extends BaseDelegateComponent<Record<string, unknown>>,
+    KEY extends keyof T['__resolve'],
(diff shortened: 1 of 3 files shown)


jBernavaPrah:
Update the PR with a fix to improve the typing when used with arrays of type. 

This last change introduces a breaking change because of the generic ARGS of Jovo.$resolve needs to be compatible with a tuple and not with an array.

So now, if users use only BaseComponent, typescript will trigger errors when multiple parameters will be used in the $resolve. Es: 
```typescript 
this.$resolve('event', "resolve1", "resolve2") // The resolve2 will be highlight by typescript as error
```

Happy to improve and remove this breaking change if someone has some suggestions!

PR #1569, opened 2023-07-04
:sparkles: Add alexa task support

sadlowskij:
## Proposed Changes

This PR should simplify working with [tasks for the alexa platform](https://developer.amazon.com/en-US/docs/alexa/custom-skills/understand-tasks.html). It instanciates a `$alexa.task` instance on the alexa instance that helps access the task properties. As the task name in the request contains a skillId before the actual task name, this helps to ignore the skill id and gives a method for checking just the actual name. 

Also it includes a new `onTask` static function for the `AlexaHandles` class, that makes setting up a handler for a specific task with a (optional) specific version quite easy:

```typescript
import { BaseComponent, Component, Global, Handle } from '@jovotech/framework';
import { LoveHatePizzaComponent } from './LoveHatePizzaComponent';
import { AlexaHandles } from '@jovotech/platform-alexa';

@Global()
@Component()
export class GlobalComponent extends BaseComponent {
  LAUNCH() {
    return this.$redirect(LoveHatePizzaComponent);
  }

  @Handle(AlexaHandles.onTask('Reminder', 1))
  reminderTask(): Promise<void> {
    const task = this.$alexa?.task.getTask();
    const taskName = this.$alexa?.task.name;
    const taskVersion = this.$alexa?.task.version;
    const inputs = this.$alexa?.task.input
    console.log(task, taskName ,taskVersion ,inputs)
    throw new Error('to be impl')
  }
  
}
```

If the code side is accepted, I can also work on docs and or tests, just let me know!

## Types of Changes

## Checklist

### DIFF
--- a/platforms/platform-alexa/docs/README.md
+++ b/platforms/platform-alexa/docs/README.md
@@ -631,6 +631,50 @@ This would result in the following output template:
 }
 ```
 
+### Custom Tasks
+
+You can handle [custom tasks](https://developer.amazon.com/en-US/docs/alexa/custom-skills/understand-tasks.html) with the following features:
+
+- `this.$alexa.task` property that offers helper methods and properties to access task related values
+- `AlexaHandles.onTask()` for the [`@Handle` decorator](https://www.jovo.tech/docs/handle-decorators) to accept task requests
+
+Here is an example:
+
+```typescript
+import { AlexaHandles } from '@jovotech/platform-alexa';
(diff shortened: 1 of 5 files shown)


aswetlow (approved):

jankoenig (approved):
Thank you @sadlowskij 🎉 

I've added some short docs that should be enough for now

PR #1615, opened 2024-01-30
:bug: Handle empty AskError detail object

sadlowskij:
## Proposed Changes

Sometimes the `detail` property in this case can be just an empty object. This happens for example, if I try to deploy while the build is still going on. It throws a 409 with an empty `detail`:
image

This will lead to an error, trying to access `payload.detail.response.message` and will cause an error message to be printed in the console that gives no information about what actually went wrong with the upload:
image

With the provided fix in place, it looks like this:
image

## Types of Changes

## Checklist


sadlowskij:
I forgot that this changed since my former PR, I will address this in a new branch/PR :)

PR #1609, opened 2024-01-02
:bug: Add missing non JSON ask error handling

sadlowskij:
## Proposed Changes
Provoking an error in alexa ask deployment sometimes results in an error in the `getAskError` function.

Example:
Using a non existent ask profile like so:

```
new AlexaCli({askProfile: 'does-not-exist', locales: {en: ['en-US']}})
```
This leads to the `stderr`: 

```
Command failed: ask smapi create-upload-url -p does-not-exist
Debugger listening on ws://127.0.0.1:42595/25792652-dbcc-412d-bf9d-4db5fcbd8620
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
Debugger listening on ws://127.0.0.1:39601/8953403c-efec-4e66-a3c0-89661f15f337
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
[Error]: Can't resolve profile [does-not-exist] as it doesn't exist. Please run "ask configure --profile does-not-exist" first.
```

The current version of `getAskError` tries to parse the substring after `[Error]`, in this case this text `Can't resolve profile [does-not-exist] as it doesn't exist. Please run "ask configure --profile does-not-exist" first.`. As this clearly is no JSON object/array. This `JSON.parse` throws an error resulting in this being displayed after running `jovo deploy:platform alexa`:

```
🚀 Deploying Alexa Skill
  ✖ Uploading skill package
                                                                                
x Error: --------------------------------------------------------------------------------
›                                                                                 
› Message:
›  Unexpected token C in JSON at position 0
›                                                                                 
› Module:
›  JovoCliCore
›                                                                                 
›                                                                                 
› If you think this is not on you, you can submit an issue here: https://github.com/jovotech/jovo-cli/issues.
```

This is why I built a block to catch the error from `JSON.parse`, which creates the `JovoCliError` in a different way, leading to the actual helpful error message being displayed after running the jovo deploy, in this case:

```
🚀 Deploying Alexa Skill
  ✖ Uploading skill package
                                                                                
x Error: --------------------------------------------------------------------------------
›                                                                                 
› Message:
›  smapiCreateUploadUrl: Can't resolve profile [does-not-exist] as it doesn't exist. Please run "ask configure --profile does-not-exist" first.
›  
›                                                                                 
› Module:
›  AlexaCli
›                                                                                 
›                                                                                 
› If you think this is not on you, you can submit an issue here: https://github.com/jovotech/jovo-cli/issues.
```

[...]

PR #1616, opened 2024-01-30
:bug: Handle empty AskError detail object

sadlowskij:
## Proposed Changes

Sometimes the `detail` property in a ask error (e. g. of a deploy) can be just an empty object. This happens for example, if I try to deploy while the build is still going on. It throws a 409 with an empty `detail`:
image

This will lead to an error, trying to access `payload.detail.response.message`. This error then is being caught [here](https://github.com/jovotech/jovo-framework/blob/v4/dev/platforms/platform-alexa/src/cli/utilities.ts#L114) which leads to the whole json bein output, instead of the message being retrieved:
image

With this fix it will correctly be retrieved:
image

I worked on this before the other change to [getAskError was merged](https://github.com/jovotech/jovo-framework/pull/1609/commits/37ec7f423adf504b3e7c35dcdd1959f449da173e) (which catches errors in the ask error retrieval process). Before this merge, the error was more severe as it completely dropped the actual error message, now it is more of a question of how it is displayed. However this still throws an error currently in a place where it is avoidable.

## Types of Changes

## Checklist

### DIFF
--- a/platforms/platform-alexa/src/cli/utilities.ts
+++ b/platforms/platform-alexa/src/cli/utilities.ts
@@ -161,7 +161,7 @@ function getViolations(payload: any): string {
     }
   }
 
-  if (payload.detail) {
+  if (payload.detail?.response?.message) {
     violations = payload.detail.response.message;
   }
   return violations;


aswetlow (approved):

jankoenig (approved):
Thank you @sadlowskij!

PR #1617, opened 2024-01-31
:bug: Fix missing skillId issue for async Alexa deploys

sadlowskij:
## Proposed Changes

Using the `--async` flag for alexa deploys calls `getImportStatus` right after deploying the skill. My testing showed that this leads to the response not being complete, see example:

image

As you can see the result here only contained `status`, while the `skill` property is not there. As Jovo is trying to access `status.skill.skillId`, this leads to the error in the screenshot.

I found out that this seems to be dependent of the time passed since the import. Waiting a moment before requesting the import status fixes this issue, see this image that shows the `status` result:
image

This is only relevant for `--async` deploys, due to non async deploys ignore `IN_PROGRESS` responses already.

I was not sure whether to put this line of code before the `getImportStatus` call as it is now, or inside `getImportStatus`. I think it's a matter of preference, feel free to change or request changes if necessary.

## Types of Changes

## Checklist

### DIFF
--- a/platforms/platform-alexa/src/cli/hooks/DeployHook.ts
+++ b/platforms/platform-alexa/src/cli/hooks/DeployHook.ts
@@ -197,6 +197,9 @@ export class DeployHook extends AlexaHook<DeployPlatformEvents> {
 
       this.$context.alexa.importId = importId;
 
+      // requesting import status right away sometimes leads to no skill info being returned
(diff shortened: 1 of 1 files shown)


aswetlow (approved):

jankoenig (approved):
Thank you @sadlowskij!

PR #1629, opened 2024-04-15
:bug: Removed false positive duplicate global handler issue

sadlowskij:
## Proposed Changes

This commit stops the same handler to be added to `globalHandlerMap` twice. This lead to an error, where it reported a duplicate handler even though it was the same handler twice. This can happen with intentMaps, e. g. you have an intentMap like this:

```typescript
  routing: {
    intentMap: {
      'AMAZON.HelpIntent': 'HelpIntent',
    }
  },
```

and a handler like this:
```typescript
  @Handle({global: true, intents: ['HelpIntent', 'AMAZON.HelpIntent']})
  help(): Promise<void> {
    throw new Error('just a test');
  }
```

I am aware that in most cases, you should just not use `AMAZON.HelpIntent` in the `intents` array and it works just fine. However, if you are using one and the same component in multiple projects that might have different intentmaps, this might actually make things easier. Also, even though this might not be the best implementation, I don't think it was intended to treat the same handler as duplicate in this specific scenario in the first place. 

## Types of Changes

## Checklist

### DIFF
--- a/framework/src/plugins/RouterPlugin.ts
+++ b/framework/src/plugins/RouterPlugin.ts
@@ -95,7 +95,10 @@ export class RouterPlugin extends Plugin<RouterPluginConfig> {
             if (!globalHandlerMap[mappedIntentName]) {
               globalHandlerMap[mappedIntentName] = [];
             }
-            if (!handlerMetadata.hasCondition) {
+            if (
+              !handlerMetadata.hasCondition &&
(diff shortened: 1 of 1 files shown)


aswetlow (approved):

jankoenig (approved):
Thank you @sadlowskij!

PR #1637, opened 2024-05-30
✨ Add verify and certify commands to alexa cli plugin

sadlowskij:
## Proposed Changes
This gives the developer an easy way to validate and certify their alexa skill by utilizing the information already present in the project configuration (like skillId, askProfile and so on). This is especially helpful if the project has many skills (we have one jovo app handling over 40 skills), where doing this manually is quite tedious.

I was not really sure about how to name this. The alexa cli so far has no commands, only hooks. If this is interesting to you, would this have to be platform independent, like `jovo validate:platform alexa` instead of how it is currently `jovo validate:alexa`?

If some of the information should be different, or the project config is not complete, there is the possibility to override those, however in the context of jovo, I think it will be most useful when it takes most information from jovo, e. g.

`jovo certify:alexa --stage dev`

I also included some documentation and heavily tested these two commands with all kinds of falg / configs, it seems to work great in my tests!

Please let me know, if you think this could be useful for Jovo!

## Types of Changes

## Checklist


aswetlow (approved):

aswetlow:
Wow, thank you!

jankoenig (approved):
This looks great, thank you @sadlowskij!

PR #1638, opened 2024-06-04
:sparkles: Add StartConnectionOutput for Alexa

sadlowskij:
## Proposed Changes

This commit adds a convenient way to add skill connections, that have not already been covered by the multiple Output classes already provided. This is especially helpful, if you want to target a specific Task (e. g. in a specific skill), as the already provided output classes only covered predefined tasks. 

The main new class added here is the `StartConnectionOutput`, which is a the most basic StartConnection Template, that should cover all types of StartConnection Directives the developer needs. Therefore I also refactored the already existing Outputs to use this more generic Output, to make these classes a bit more concise. 

I tested the new class heavliy as well as made sure, that the refactored classes still work the same. In my tests, everything worked perfectly.

I also added documentation with links to the alexa developer documentation to explain how to use this new feature!

## Types of Changes

## Checklist


sadlowskij:
In case this feature is interesting, but the refactoring of the old classes is not desired, I could also just revert the refactoring part and leave those classes as they were!

aswetlow (approved):

jankoenig (approved):
Thank you!

